PR #914, opened 2022-03-29
Fix dodge direction

socolio:
Movement input that was calculated by keybind reading can be modified by event listeners from other mods. It will be better to use player movement input as a source for dodge direction. Just more compat with other mods!


Yesssssman:
Hello, Can I know the name of the representative mod that this change will affect?

socolio:
This fix will add compat with https://www.curseforge.com/minecraft/mc-mods/better-third-person

Yesssssman:
Alright thanks for your attribution!

PR #1115, opened 2022-08-07
Fix usage of AnimationRegistryEvent for other mods

Wehavecookies56:
This moves the animation registration to the FMLCommonSetupEvent so it's possible for other mods to register a listener to the event, without this the event is posted before other mods can register a listener as the dependency load order doesn't do anything when they initially load in parallel. I also moved the Skill registry as that relies on the animations being registered first and the SkillRegistryEvent should work for other mods too with this change.

As far as I can tell nothing is broken by this change and testing with Kingdom Keys I was able to register a new animation.


Yesssssman (approved):
Well... it worked fine when I tested it with Weapons of Minecraft (Epic Fight Addon). But yeah, after changing to this code still works fine and it looks better. Thanks!

PR #1349, opened 2023-07-13
Support 1.19.2

ISNing:
Tested. But there are still two "FIXME" where I don't understand need to be reviewed

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,7 +25,7 @@ archivesBaseName = 'epicfight'
 java.toolchain.languageVersion = JavaLanguageVersion.of(17)
 
 minecraft {
-    mappings channel: 'official', version: '1.18.2'
+    mappings channel: 'official', version: '1.19.2'
     accessTransformer = file('src/main/resources/META-INF/accesstransformer.cfg')
 
     runs {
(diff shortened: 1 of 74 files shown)


ISNing:
What's more, although I think this may should pr to branch 1.19.2, as for it didn't exist, I don't know how to do that, sorry~

ISNing:
[epicfight-18.3.8.zip](https://github.com/Yesssssman/epicfightmod/files/12260089/epicfight-18.3.8.zip)
Here's artifacts I built, remember to change it's extension to `.jar`

Manwe03:
please, make the 1.19.2 branch public in curseforge

ISNing:
> please, make the 1.19.2 branch public in curseforge

Sorry, but I don't think I have enough free time to maintain it
You can just use the jar I have already built, it's been tested and no obvious bug has been found

Kasualix:
Can be closed now. ~It seems like their GitHub is just for show to external contributors.~

ISNing:
> Can be closed now. ~It seems like their GitHub is just for show to external contributors.~

You're right, maybe the have never expected someone to contribute to this.....

MetalKnight56:
> Can be closed now. ~It seems like their GitHub is just for show to external contributors.~

I don't get what you mean in "just for show" ? We decide what we accept or not, if we didn't accept the request back then it means that we don't want that contribution for now. We thank the author for helping us on this, but we have our own contributors that made the port in our own models. If you don't comprehend that, i don't think you are in the right place.

PR #1409, opened 2023-10-21
Allow blank issues

Kasualix:
# Why
One thing you guys seem to be forgetting is that GitHub Issues can be created for feature requests/suggestions as well.

There has been a feature request template idk why it's removed, but at least I don't think it's a good idea to fill in the bug report with things that are completely useless for Suggestion/Feature Requests

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,2 +1,2 @@
-blank_issues_enabled: false
+blank_issues_enabled: true
 contact_links:


MetalKnight56:
> # Why
> One thing you guys seem to be forgetting is that GitHub Issues can be created for feature requests/suggestions as well.
> 
> There has been a feature request template idk why it's removed, but at least I don't think it's a good idea to fill in the bug report with things that are completely useless for Suggestion/Feature Requests

We handle suggestions in our discord server, its easier and we're more oftenly active there. We do accept suggestions from here too, but we mainly use github for bug reports.

MetalKnight56:
As of this commit: https://github.com/Yesssssman/epicfightmod/commit/3cd4e802bf770272c197b047a68b6c09b191eb43 , i've enabled blank issues to be created.
Also created a new issue template for those wanting to make a suggestion.

PR #1473, opened 2023-12-04
[1.20.1] Add AzureLib/AzureLib Armor Support

AzureDoom:
Adds support AzureLib/AzureLib Armor armors in 1.20.1 (as it's based on GeckoLibs 4.x branch).


EchoEllet:
Thank you for the PR!

Looks solid, however, it depends on some internal AzureLib APIs, such as `mod.azure.azurelib.common.internal.client.util.RenderUtils`, which has been removed in **AzureLib 3** (#2279) and not documented in the WIKI or in Javadoc.

> as it's based on GeckoLibs 4.x branch)

This seems to be no longer the case.

As a workaround, I disabled **AzureLib 2** compatibility modules if the mod's version is 3 (or greater) in this PR #2272 and added some logging for debugging purposes. This was done as we're unable to port in a reasonable time frame for now, as it would require looking into the AzureLib's commit history and the differences.

If there are any documents on how we can migrate, or if you're aware of the recent changes, please consider letting me know.

PR #1533, opened 2024-02-01
Make mixin code less invasive for compatability reasons.

eehunter:
This should theoretically fix a conflict with Incubus Core that occurs when the two are used with Sinytra Connector.


Yesssssman:
Thank you

PR #1600, opened 2024-04-09
fix stun immunity and parry continuously

Cyber2049:
Fixed stun immunity not working on short stun type


Yesssssman:
Great work!

PR #1836, opened 2024-10-09
fix livingAttackEvent

Cyber2049:
make sure the animation of damagesource won't change if this  orignial damagesource is belongs to epicfight.


Yesssssman:
ty!

PR #1939, opened 2025-03-24
Update WearableItemLayer.java to support BetterNether

Raiiiden:
Fixed BetterNether crash


Yesssssman:
image

I think I'll just wrap the execution with try-catch lol, We don't exactly know what causes this bug and it could have more issues not only with armor stand.

PR #1955, opened 2025-04-18
Add a try-catch statement wrapping loadAnimationClip for debugging purposes

Forixaim:
One of the exceptions I've read was a weird one.
image
This exception message was almost undebuggable since the exception didn't give what animation file caused the error.

So I've wrapped the code that the exception was thrown from with a try-catch statement that rethrows the exception except with a message that points what animation file caused the error. That way, it is significantly easier to find the erroring animation and apply fixes.


Forixaim:
image


Here is the extra message that is printed alongside the exception message.

PR #1957, opened 2025-04-21
Added comparison of the Timestamp Array and Transform Matrix Array alongside improving debuggability for datapacks by pointing what error it is

Forixaim:
This can further improve debuggability to see whether the timestamp array is too large or too small. Which gives specifics on whether the animation transform sheet is too long or not


Forixaim:
will reopen to reflect updated commits

PR #1998, opened 2025-06-15
Disable mouse inputs when target lock-on is active

Exopandora:
This PR fixes a bug where mouse inputs remain enabled when target lock-on is active, causing visual jitter in third person. You can view an example video of the bug by @JustDudeIt123 [here](https://www.youtube.com/watch?v=6VpznTb3RNY). They originally reported the bug [here](https://github.com/Exopandora/ShoulderSurfing/issues/310).


Yesssssman:
Great job, thanks!

PR #2035, opened 2025-07-30
Updated Skill Charging Documentation and added Holdable Skill alongside many tweaks.

Forixaim:
Changes: Guard Skill can now be rebound, and it's now classified as a HoldableSkill
Fixed typo on ControllEngine -> ControlEngine
New HoldableSkill with complete documentation.
Revamped ChargeableSkill Documentation


Forixaim:
Added a dummy class to avoid breaking addons until 20.12.x

Issue #1297, opened 2023-03-28
[Bug]: Controllable support


911653534:
### Have you checked if a similar issue is already reported by someone else?

- [ ] I checked there are no similar issues have been reported.

### Have you read the support policy?

- [X] I read it and I accept the policy.

### Are you using the latest Epic Fight and recommended Forge version?

- [X] I checked I'm using latest Epic Fight and recommended Forge version.

### Is this issue related to mod incompatibility?

- [ ] This is a mod compatibility issue and I'm aware of the problem.

### Minecraft Version

1.18.2

### What happened?

i'd like to replace the left mouse button with some other key, so that i can use my gamepad conveniently. the main problem is that i can't attack in fight mode when i also use the mod named Controllable

ChaoticPhamton:
it's the main button on mc, like the vanilla attack, the button that appears in controls are for the skill, have you tried it that way yet?

911653534:
Thank you very much for seeing my message, but when I use the Controllable mod and use the corresponding attack button, the behavior produced is the original attack behavior, not the special style attack in the Epic Fight mod. Maybe my description is not very clear, but I have seen other players suffer from the same problem. Can you try the Controllable mod？

911653534:
If the attack key in attack mode of Epic Fight mod can be displayed and modified in the key settings interface, I think it would be a great modification solution. I'm afraid my explanation is not clear, and here is the result picture after the change. Thank you for your reply again.
![1](https://user-images.githubusercontent.com/75149084/228583129-d1fb9957-08ec-4e74-9f1f-0d828c39f751.png)

MetalKnight56:
> 

Controllable compatibility is not available yet, make sure to remove that mod for now and follow our progress as our next update should cover these bugs.

MetalKnight56:
I'll be closing this issue when the update is released. Lets leave this open to help me remember to inclusde this on the update note.

Blues7a:
I hope this is fixed soon

MetalKnight56:
Issue has been fixed as of the latest commits
https://github.com/Yesssssman/epicfightmod/compare/e963d283e2ec...35e8aa9ea4ba

EchoEllet:
> Issue has been fixed as of the latest commits [e963d28...35e8aa9](https://github.com/Epic-Fight/epicfight/compare/e963d283e2ec...35e8aa9ea4ba)

[...]

PR #2112, opened 2025-10-14
feat: add support for Shoulder Surfing Reloaded mod

### LINKED ISSUE #2111: Shoulder Surfing mod does not detect Epic Fight custom Attack keybind
### Background

The **Shoulder Surfing Reloaded** mod includes a camera decoupling feature, allowing the player to rotate the camera 360° without rotating their controlled character. When the vanilla **Attack** key is pressed, Shoulder Surfing automatically detects it and rotates the player to face the crosshair target as needed to attack enemies.

---

### Issue

The **Epic Fight** mod added a custom **Attack** keybind in [this commit](https://github.com/Epic-Fight/epicfight/compare/e963d283e2ec...35e8aa9ea4ba#diff-9dc988ba2888a8eece19c042d1412fc79025c1c46370f5c0b5974a597c56ba5bL76)  
to support controller mods (more details: [#1297](https://github.com/Epic-Fight/epicfight/issues/1297)).  
However, Shoulder Surfing Reloaded **cannot detect this custom keybind** if the vanilla Attack key differs from the Epic Fight Attack key, which is usually the case when using a controller.  

This means players must always **disable the camera decoupling feature** to make the mod playable with a controller, and are forced to use the **same Attack key** as in vanilla and Epic Fight.

<details><summary>Current behavior video</summary>
<p>

https://github.com/user-attachments/assets/b4488859-c5bb-4ad2-b72b-fe4ac35a1214

Refer to the video in the PR #2112

</p>
</details> 

---

### Solution

Registers a Shoulder Surfing mod plugin (using the supported API) to force camera coupling while attacking.

> [!TIP]
> Fixed in the PR #2112.

<details><summary>Alternative workaround (less preferred)</summary>
<p>

Works around the issue by overriding the [`ShoulderSurfingImpl.isAttacking`](https://github.com/Exopandora/ShoulderSurfing/blob/7f0df83beb4f7158810e188150eb7e9812981529/common/src/main/java/com/github/exopandora/shouldersurfing/client/ShoulderSurfingImpl.java#L188-L191) method with a mixin, allowing the Shoulder Surfing mod to support the Epic Fight Attack keybind as well.

</p>
</details>

### LINKED ISSUE #2113: Shoulder Surfing mod doesn't turn player when attacking by default
### Background

The **Shoulder Surfing Reloaded** mod includes a camera decoupling feature, allowing the player to rotate the camera 360° without rotating their controlled character. When attacking, Shoulder Surfing automatically rotates the player to face the crosshair target as needed to attack enemies **but only if there is a target** (default).

---

### Issue

The mechanics of turning only if there is a target while attacking works pretty well on vanilla but unplayable with Epic Fight, even on mouse and keyboard.

<details><summary>Current behavior video</summary>
<p>

https://github.com/user-attachments/assets/f35e52eb-cad4-45fe-bbd0-742d8f5107d3

</p>
</details> 

---

### Solution

Registers a Shoulder Surfing mod plugin (using the supported API) to force camera coupling while attacking.

> [!TIP]
> Fixed in the PR #2112.

### LINKED ISSUE #2114: Player doesn't turn in Shoulder Surfing when holding Epic Fight skills
### Background

The **Shoulder Surfing Reloaded** mod includes a camera decoupling feature, allowing the player to rotate the camera 360° without rotating their controlled character. When performing certain actions, such as throwing eggs or holding an egg in the main hand, Shoulder Surfing temporarily forces camera coupling to aim.

---

### Issue

However, when charging an Epic Fight skill such as **Demolition Leap** (or holding any Epic Fight skill), it does not force camera coupling.

<details><summary>Current Behavior Video</summary>
<p>

https://github.com/user-attachments/assets/c6b3e27e-0912-419c-b13c-58b8a9babb5c

</p>
</details> 

### Solution

Registers a Shoulder Surfing mod plugin (using the supported API) to force camera coupling while holding any skill, including but not limited to Demolition Leap, to support other mods as well.

> [!TIP]  
> Fixed in PR #2112.

EchoEllet:
**See the related issue or JavaDoc comment in the PR code for more details (#2111).**

### Related Issues

- *Fix #2111*
- *Fix #2113*
- *Fix #2114*
- *Related #1297*
- *Related Commit https://github.com/Epic-Fight/epicfight/compare/e963d283e2ec...35e8aa9ea4ba#diff-9dc988ba2888a8eece19c042d1412fc79025c1c46370f5c0b5974a597c56ba5bL76*
- *Related https://github.com/Exopandora/ShoulderSurfing/issues/345*

### Videos

Tested with both [Controllable](https://www.curseforge.com/minecraft/mc-mods/controllable) and [Controlify](https://modrinth.com/mod/controlify) mods (should work with any controller mods, the fix is independent).

<details><summary>Before</summary>
<p>

https://github.com/user-attachments/assets/b4488859-c5bb-4ad2-b72b-fe4ac35a1214

</p>
</details> 

<details><summary>After</summary>
<p>

Steel whirlwind works well now:

https://github.com/user-attachments/assets/f460c3fc-b81e-47e7-9bf7-d998147cbaa0

Demolition leap is now compatible too:

https://github.com/user-attachments/assets/f51f1a94-eca1-4510-b715-5d2228be92e7

With controllable mod:

https://github.com/user-attachments/assets/53235a3b-5124-49b1-b096-0adc4dc78ab1

With Controlify mod:

https://github.com/user-attachments/assets/02cb0ca5-db95-4a7a-a7c1-dbc813b1a8ca

</p>
</details> 

Refer to [Shoulder Surfing API docs](https://github.com/Exopandora/ShoulderSurfing/wiki/API-Documentation-Plugins#creating-a-plugin-class)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -136,6 +136,18 @@ dependencies {
     
     // First-person compatibility (optional) (Warn: this mod file will cause crash, as it references obfuscated names)
     compileOnly "local:firstperson:2.5.0"
+
+    // Shoulder Surfing Reloaded compatibility (optional)
+    compileOnly "curse.maven:shoulder-surfing-reloaded-243190:6993797" // API only (https://www.curseforge.com/minecraft/mc-mods/shoulder-surfing-reloaded/files/6993797)
+    localRuntime "curse.maven:shoulder-surfing-reloaded-243190:6993795" // Full mod at runtime (https://www.curseforge.com/minecraft/mc-mods/shoulder-surfing-reloaded/files/6993795/)
+
+    // Uncomment to test Controllable compatibility
+    /*localRuntime "curse.maven:controllable-317269:6850105"
+    localRuntime "curse.maven:framework-549225:6531439"*/
+
+    // Uncomment to test Controlify compatibility
+    /*localRuntime "curse.maven:controlify-835847:6674779"
+    localRuntime "curse.maven:yacl-667299:6662857"*/
 }
 
 var generateModMetadata = tasks.register("generateModMetadata", ProcessResources) {
--- a/src/main/java/yesman/epicfight/compat/ShoulderSurfingCompat.java
+++ b/src/main/java/yesman/epicfight/compat/ShoulderSurfingCompat.java
@@ -1,0 +1,72 @@
+package yesman.epicfight.compat;
+
+import com.github.exopandora.shouldersurfing.api.callback.ICameraCouplingCallback;
+import com.github.exopandora.shouldersurfing.api.plugin.IShoulderSurfingPlugin;
+import com.github.exopandora.shouldersurfing.api.plugin.IShoulderSurfingRegistrar;
+import net.minecraft.client.Minecraft;
+import net.neoforged.api.distmarker.Dist;
+import net.neoforged.api.distmarker.OnlyIn;
+import yesman.epicfight.client.ClientEngine;
+import yesman.epicfight.client.events.engine.ControlEngine;
+import yesman.epicfight.client.input.EpicFightKeyMappings;
+import yesman.epicfight.client.world.capabilites.entitypatch.player.LocalPlayerPatch;
+
+/**
+ * Background:
+ * The Shoulder Surfing Reloaded mod includes a camera decoupling feature, allowing the player to rotate the camera 360°
+ * without rotating their controlled character. When the vanilla Attack key is pressed,
+ * Shoulder Surfing automatically detects it and rotates the player to face the crosshair target as needed to attack
+ * enemies.
+ * <p>
+ * <p>
+ * The Epic Fight mod added a custom Attack keybind in this commit: <a href="https://github.com/Epic-Fight/epicfight/compare/e963d283e2ec...35e8aa9ea4ba#diff-9dc988ba2888a8eece19c042d1412fc79025c1c46370f5c0b5974a597c56ba5bL76">...</a>
+ * to support controller mods (more details: <a href="https://github.com/Epic-Fight/epicfight/issues/1297">...</a>).
+ * However, Shoulder Surfing Reloaded cannot detect this custom keybind if the vanilla Attack key
+ * differs from the Epic Fight Attack key, which is usually the case when using a controller.
+ * This means players must always disable the camera decoupling feature to make the mod playable with a controller
+ * and are forced to use the same Attack key as in vanilla and Epic Fight.
+ * <p>
+ * <p>
+ * This Shoulder Surfing plugin works around the issue by bypassing the
+ * <a href="https://github.com/Exopandora/ShoulderSurfing/blob/7f0df83beb4f7158810e188150eb7e9812981529/common/src/main/java/com/github/exopandora/shouldersurfing/client/ShoulderSurfingImpl.java#L188-L191">
+ * ShoulderSurfingImpl.isAttacking
+ * </a> method,
+ * allowing the Shoulder Surfing mod to support the Epic Fight Attack keybind as well.
+ * <p>
+ * <p>
+ * This plugin always forces the player to turn when attacking, regardless of the Shoulder Surfing "player.turning" config.
+ * By default, Shoulder Surfing only turns the player when a target is detected ("REQUIRES_TARGET"),
+ * which works for vanilla mechanics but is incompatible with Epic Fight.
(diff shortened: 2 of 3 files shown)


build.gradle
```diff
@@ -136,6 +136,13 @@ dependencies {
     
     // First-person compatibility (optional) (Warn: this mod file will cause crash, as it references obfuscated names)
     compileOnly "local:firstperson:2.5.0"
+
+    // Shoulder Surfing Reloaded compatibility (optional)
+    implementation "curse.maven:shoulder-surfing-reloaded-243190:6222260"
+
+    // Uncomment to test controller compatibility
+    /*localRuntime "curse.maven:controllable-317269:6850105"
+    localRuntime "curse.maven:framework-549225:6531439"*/
```
EchoEllet: I think it's worth adding these instead for testing:

```kotlin
localRuntime "curse.maven:controlify-835847:6674779"
localRuntime "curse.maven:yacl-667299:6662857"
```

Controlify is more modern, has supported API, and already works with most Epic Fight basic features such as dodging in the correct direction (even with Shoulder surfing mod installed). Also, it supports setting multiple keybinds to the same controller button, which something does not work on Controllable AFAIK.

src/main/java/yesman/epicfight/mixin/shoulder_surfing/MixinShoulderSurfingImpl.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+@OnlyIn(Dist.CLIENT)
+@Mixin(ShoulderSurfingImpl.class)
+public class MixinShoulderSurfingImpl {
+    @Redirect(
+            method = "isAttacking(Lnet/minecraft/client/Minecraft;)Z",
+            at = @At(
+                    value = "INVOKE",
+                    target = "Lnet/minecraft/client/KeyMapping;isDown()Z"
+            )
```
EchoEllet: I think there might be an alternative way to fix this issue for all mods without modifying or depending on the Shoulder Surfing mod at all.

The idea is to mix in on any `isDown` call, and if the key mapping is the vanilla attack key and the original return value of `isDown` was `false`, and the vanilla attack key is not the same as the Epic Fight attack key, we could try checking the Epic Fight attack key instead and return `true`. Otherwise, we ignore the injection.

I'm not exactly sure if this solution could introduce unpredictable bugs with other mods, so I decided to hardcode the solution and fix it only for Shoulder Surfing. It might be worth testing this with other third-person mods such as Better Third Person, and if the solution proves to be more universal with no noticeable issues, then it might be worth considering. What do you think?

src/main/java/yesman/epicfight/compat/ShoulderSurfingCompat.java
```diff
@@ -0,0 +1,55 @@
...
+public class ShoulderSurfingCompat implements IShoulderSurfingPlugin {
+    @Override
+    public void register(IShoulderSurfingRegistrar registrar) {
+        registrar.registerCameraCouplingCallback(new ForceCameraCouplingWhenAttackingCallback());
+    }
+
+    private static class ForceCameraCouplingWhenAttackingCallback implements ICameraCouplingCallback {
+        @Override
+        public boolean isForcingCameraCoupling(Minecraft minecraft) {
+            return ControlEngine.isKeyDown(EpicFightKeyMappings.ATTACK) || minecraft.options.keyAttack.isDown();
```
EchoEllet, line 55: I'm not entirely sure whether we should use `ControlEngine.isKeyDown` instead of the vanilla `KeyMapping.isDown` for only `EpicFightKeyMappings.ATTACK` or for the vanilla attack key too. I used Epic Fight's `ControlEngine.isKeyDown` because it calls `KeyMapping.isDown` anyway.

staffordpreston:
I got 1 more problem what is this mod bro I can't find it anywhere in my pack 
image

EchoEllet:
@staffordpreston This PR is for adding support Shoulder Surfing support, [Epic Fight - ExCap](https://www.curseforge.com/minecraft/mc-mods/ex-cap) is unrelated mod. This mod requires a newer version of Epic Fight, so you need to update Epic Fight, though, there are many Epic Fight addons that still require older versions of Epic Fight.
This isn't an Epic fight issue.

EchoEllet:
I understand that you want help fixing this issue, but this isn't the right place for that. GitHub issues and PRs are intended for tracking and progressing the project. Please post your question on the Epic Fight Discord instead.

PR #2120, opened 2025-10-19
feat(shoulder-surfing): support Epic Fight attack key and always turn when attacking for 1.20.1

EchoEllet:
Cherry-pick PR #2112, originally made for 1.21.1, into the 1.20.1 branch.


EchoEllet:
BTW, I'm collaborating with the Shoulder Surfing Reloaded mod author (https://github.com/Exopandora/ShoulderSurfing/issues/345), so we can add support for a better callback that improves compatibility. Once that is done, I will send a PR for Epic Fight to use the new callback for both 1.20.1 and 1.21.1 (should be a minor change). Users will need to update the Shoulder Surfing mod, which is an easy fix.

PR #2121, opened 2025-10-19
Remove mod JAR files in libs/ directory by using Modrinth maven

EchoEllet:
Including mod JAR files or any large binary files in a Git repository is inefficient because these mods often introduce breaking changes, and we frequently need to port and update versions. Using these mods from [Modrinth Maven](https://support.modrinth.com/en/articles/8801191-modrinth-maven) minimizes maintenance and work.  

Optionally, this also make use of Gradle 6.2+ `exclusiveContent` for better isolation, filtering, and build performance by minimizing network calls. This may seem minor, but it is useful when using GitHub workflows that run on every commit, as is the case in the `1.21.1` branch.

IMO, it is better to prefer Modrinth Maven over CurseForge whenever possible for the following reasons:

* **Better support for classifiers:** Modrinth natively supports classifiers such as `sources` and `javadoc`, allowing developers to view Epic Fight’s source code directly in their IDE without additional setup. This makes exploring the Epic Fight APIs much easier, without having to navigate to the GitHub repository for a specific branch, commit, or file.
* **Human-readable versioning:** CurseForge enforces numeric version IDs (e.g., `7106604`), while Modrinth supports readable version strings (e.g., `21.12.5`) that can be used in `gradle.properties` or `gradle/libs.versions.toml`.
* **Open and developer-friendly platform:** Modrinth is community-driven, open-source, provides a public API without requiring an API key, and fully supports third-party integrations without restrictions.  
  It also offers a **cleaner UI**, **faster performance**, and is **Minecraft-focused**, making it a better fit overall.

I kept the existing lines that use the CurseForge Maven repository if that is your preference.


build.gradle
```diff
@@ -78,12 +78,17 @@ repositories {
...
+
+    exclusiveContent {
+        forRepository { maven { url "https://cursemaven.com" } }
+        forRepositories(fg.repository)
+        filter { includeGroup "curse.maven" }
+    }
+
+    exclusiveContent {
+        forRepository { maven { name = "Modrinth"; url = "https://api.modrinth.com/maven" } }
+        forRepositories(fg.repository)
```
EchoEllet, line 90: FYI: On NeoForge 1.21.1, we just need to remove the `forRepositories(fg.repository)` line and that's all.

PR #2122, opened 2025-10-19
Refactor Epic Fight Input System

### LINKED ISSUE #2116: Refactor the Epic Fight Input System for low coupling and extensibility (e.g., to support controllers).
### Proof of Concept

> [!TIP]
> **Update:** Fixed in #2122, #2133 and https://github.com/Epic-Fight/epicskills/pull/5 (and other related PRs)

<details><summary>Tap to show/hide Videos</summary>
<p>

https://github.com/user-attachments/assets/fac9b206-b147-411c-9051-323ecb125233

https://github.com/user-attachments/assets/588b4418-9179-49fd-b113-f51ed2e20a2a

https://github.com/user-attachments/assets/c21e241d-998c-4748-83a9-b2c18a0eaa5a

I could still attack mobs in epic fight mode and then destroy blocks within the same controller input button. 

https://github.com/user-attachments/assets/a2abda27-a6fc-46b5-8e76-8aeeace50c3b

</p>
</details> 

<details><summary>Download this prototype version</summary>
<p>

I have uploaded the modified Epic Fight mod and Controlify for quick testing.  
This is just a **dummy experimental** version that includes changes I made while testing to better plan this issue.  
All of these changes will be reverted and reworked in the final production version:  
[prototype.zip](https://drive.google.com/file/d/1m2ic-wlDdY-OAxkjxqaibCeaoDJxovon/view?usp=drive_link)

Drag and drop the files into your `.minecraft` directory and replace the existing ones.  
The Controlify version in this ZIP was built using this PR:  
https://github.com/isXander/Controlify/pull/689

Alternatively, use [this addon](https://modrinth.com/mod/epic-fight-controlify), which uses mixin injections on Epic Fight and Minecraft to fix the compatibility issues with Controlify. This addon was made for this prototyping.

</p>
</details> 

> [!IMPORTANT]
> Hey players! 👋 If you want to show support or interest in this issue, **please just click the :thumbsup: reaction** instead of posting a comment.  
> GitHub issue comments are reserved to track technical details, so this helps keep things organized for everyone. Thanks!

### Details

This issue does not track support for a specific controller mod. Instead, it aims to perform cleanup and refactoring so that supporting any controller mod becomes much smoother and easier.  
This allows other mods—or even Epic Fight itself—to add support for controller mods with low coupling, ensuring there is exactly one class that serves as the implementation for a specific controller mod.

### API Prototype / Pseudo Code

We need common code design that centralizes all necessary fixes and changes, so updates only need to be made in one place and automatically apply everywhere::

<details><summary>Common code (protype)</summary>
<p>

```java

// FYI: This is the prototype dummy implementation,
//  I was able to come up with a much better system design:
//  https://github.com/Epic-Fight/epicfight/pull/2122

// BTW: We should maintain compatibility with existing Epic Fight addons by preserving how they use HoldableSkill.getKeyMapping.
// This ensures addons can continue registering custom key mappings for Holdable skills without requiring changes
// to the new system. The safest approach is to keep HoldableSkill.getKeyMapping as a fallback rather than deprecating
// or removing it.
public enum EpicFightInputAction {
    ATTACK,
    JUMP,
    MOBILITY,
    GUARD,
    DODGE,
    LOCK_ON,
    TOGGLE_BATTLE_MODE,
    WEAPON_INNATE_SKILL,
    MOVE_FORWARD,
    MOVE_BACKWARD,
    MOVE_LEFT,
    MOVE_RIGHT,
    SPRINT,
    CROUCH
    // ....
}

// This isn't the controller itself, but rather what provides input by abstracting any
// third-party controller mod API.
public interface IEpicFightController {
    boolean inControllerMode();

    boolean isActionJustPressed(EpicFightInputAction action);

    boolean isActionTriggered(EpicFightInputAction action);
}

// An implementation that depends on this mod: https://modrinth.com/mod/controlify
// The oldest supported Minecraft versions are NeoForge 1.20.3 and Fabric 1.19.3.
// The mod is usually ported quickly and is already available up to version 1.21.10.
public class ControlifyEpicFightController implements IEpicFightController {
    private static ControlifyApi getApi() {
        return ControlifyApi.get();
    }

    @NotNull
    private static InputBinding getInputBinding(EpicFightInputAction action) {
        final InputBindingSupplier supplier = switch (action) {
            // We can add custom Controlify bindings that could disable other bindings but just for controllers.
            case ATTACK -> EpicFightControlifyCompat.COMBAT_ATTACK;
            case VANILLA_ATTACK_DESTROY -> ControlifyBindings.ATTACK;
            case JUMP, MOBILITY -> ControlifyBindings.JUMP;
            case WEAPON_INNATE_SKILL -> ControlifyBindings.ATTACK;
        };
        return supplier.on(ControlifyApi.get().getCurrentController().get());
    }

    @Override
    public boolean inControllerMode() {
        return getApi().currentInputMode() == InputMode.CONTROLLER;
    }

    @Override
    public boolean isActionJustPressed(EpicFightInputAction action) {
        return getInputBinding(action).justPressed();
    }

    @Override
    public boolean isActionTriggered(EpicFightInputAction action) {
        return getInputBinding(action).digitalNow();
    }
}

// OR make backward compatible changes to ControlEngine
@OnlyIn(Dist.CLIENT)
public final class EpicFightControls {
    private EpicFightControls() {
    }
    private static final Options OPTIONS = Minecraft.getInstance().options;

   // We can set this to null by default and let other mods implement it to integrate their controller mod.
   private static final IEpicFightController controllerInputProvider = new ControlifyEpicFightController();

    @NotNull
    private static KeyMapping getKeyMapping(EpicFightInputAction action) {
        return switch (action) {
            case JUMP -> OPTIONS.keyJump;
            case MOBILITY -> EpicFightKeyMappings.MOVER_SKILL;
            case WEAPON_INNATE_SKILL -> EpicFightKeyMappings.WEAPON_INNATE_SKILL;
            ...
        };
    }

    // Identical to ControlEngine.isKeyDown() but with controller support
    public static boolean isActionTriggered(EpicFightInputAction action) {
        return getKeyMapping(action).isDown() || controllerInputProvider.isActionTriggered(action);
        // OR ControlEngine.isKeyDown(getKeyMapping(action))
    }

    private static boolean isUsingController() {
        return controllerInputProvider.inControllerMode();
    }

    // Identical to Epic fight ControlEngine.isKeyPressed but supports controller while still supporting NeoForge/Forge events
    public static boolean isActionJustPressed(EpicFightInputAction action, boolean eventCheck) {
            KeyMapping keyMapping = getKeyMapping(action);
    
           // The behavior is different for now between the consumeClick and jsutPressed but that will be addressed in the final version
            boolean consumes = !isUsingController() ? keyMapping.consumeClick() : controllerInputProvider.isActionJustPressed(action);
    
            if (consumes && eventCheck) {
                int mouseButton = InputConstants.Type.MOUSE == keyMapping.getKey().getType() ? keyMapping.getKey().getValue() : -1;
                InputEvent.InteractionKeyMappingTriggered inputEvent = ClientHooks.onClickInput(mouseButton, keyMapping, InteractionHand.MAIN_HAND);
    
                if (inputEvent.isCanceled()) {
                    return false;
                }
            }
    
            return consumes;
        }
    }
```

> **Note:** You might have noticed that I didn't modify the existing `ControlEngine`. Instead, I introduced `EpicFightControls` (a better name might be possible), which will be used by `ControlEngine`. This is intentional, since `ControlEngine` resides in `yesman.epicfight.client.events.engine`, which isn't part of the public API. The goal is to introduce a minimal public API that provides useful functionality rather than moving `ControlEngine` (which would be a breaking change or expose the full internal `ControlEngine`).  
> 
> The best place for `EpicFightControls` is likely in `yesman.epicfight.api.client.input`, and it must not depend on `ControlEngine`. Only `ControlEngine` depends on it to avoid tight coupling.

</p>
</details> 

Now we should use that common code across all places that violates the principles (see `General Principle` section bellow),
For example, [`PhantomAscentSkill.java`](https://github.com/Epic-Fight/epicfight/blob/9cf2e7918daa64c82d8496134c4ca3a12ced1fb5/src/main/java/yesman/epicfight/skill/mover/PhantomAscentSkill.java#L69):

```diff
- boolean jumpPressed = Minecraft.getInstance().options.keyJump.isDown();
+ boolean jumpPressed = EpicFightControls.isActionTriggered(EpicFightInputAction.JUMP);
```

Mods can now inject `EpicFightControls` via a mixin or register an event to plug in their custom logic, and now the Phantom Ascent double jump works:

<details><summary>Working double jump</summary>
<p>

https://github.com/user-attachments/assets/e8d38dd8-36fb-4d79-a881-d452a1b6aa95

</p>
</details> 

Of course, supporting controller mods without introducing dependency coupling will require more work, but I believe the effort is worthwhile given the long-term benefits in maintainability, extensibility, and overall code quality.

It's important that, while designing, we **keep compatibility with other mods in mind**.  

- **Existing mods** that use the current input system should continue to work correctly **without any changes**.  
- **New mods** can either:
  - Use the **vanilla input system**, or  
  - Adopt the **Epic Fight input abstractions**, providing a **centralized place for controller compatibility** that works across all Epic Fight mods.

### Other Advantages

This isn’t just about adding compatibility for controller mods. It also provides several key benefits:

* **Centralized logic** – All input handling is managed in one place, so bug fixes can be applied once rather than scattered across the codebase. Extract duplicated code.
* **Single responsibility** – Each system focuses on its own task, improving code organization and maintainability.  
* **Low coupling** – Reduces dependencies between systems, making future integration smoother.  
* **Easier porting** – With fewer direct references to Minecraft code, porting to new versions or platforms becomes slightly easier.

#### 1. Single-Trigger Inputs

These are inputs that should trigger **only once when the button** is pressed (e.g., opening menus, locking on a target).

#### 2. Continuous Hold Actions

Inputs that should trigger **every tick while the button is held down** (e.g., Guard).

#### 3. Long-Press Detection

Some skills (e.g., Demolition leap) require holding a button for a certain duration before triggering.

### Collaboration

<details><summary>Collaboration / Personal thoughts</summary>
<p>

I'm open to collaborate with the Epic Fight team to work on this issue, I may need to ask a few questions or code review, other than that, I'm able to take over this issue.

I think it's better to improve the API design and simplify the existing code before adding any new features or even controller mod to reduce complexity. So the current focus of this issue is not about controller support at all.

Maybe even design it in a way so that Epic Fight addons (e.g., WOM, Night fall and Sword Soaring) can be compatible too with minimal effort.

The goal isn’t to get it to just work (I’ve already done that, as shown above), but to get it working properly.

</p>
</details> 

### Related issues

- *Related #1297*
- *Related #2019*
- *Related #1547*
- *Related #417*
- *Related commit https://github.com/Epic-Fight/epicfight/compare/e963d283e2ec...35e8aa9ea4ba*

### Controller Checklist

This is a list of actions Epic Fight should take to make it easier for other mods—and Epic Fight itself—to support third-party controllers. These are the absolute minimum changes needed to support controllers.

#### General Principle

We should always:

##### **1. Avoid checking vanilla keybinds directly**, such as `Minecraft.getInstance().options.keyJump.isDown()` or `Minecraft.getInstance().options.keyJump.consumeClick()`

These checks will always return `false` when using any controller mod. Instead, we should register a controller-mod input binding and use it without directly depending on a specific controller mod.

<details><summary>Why this matters</summary>
<p>

* Controller mods disable vanilla keybinds for controllers only and replace them with input bindings designed specifically for the controller mod.
* Controller mods may also register unsupported modded keybinds to emulate the original keybind, so `isDown` might sometimes return `true` (unpredictably!). This means a workaround could be to register an Epic Fight-specific keybind that performs jump, which any controller mod would detect—but this is a hacky approach. We should avoid it and stick to the current, proper method.

</p>
</details>

#### **2. Avoid directly querying vanilla `Input` for player movement**

Instead of checking `Input.up`, `Input.down`, `Input.left`, `Input.right` directly—which only reflects keyboard input—we should introduce an abstraction to determine movement that supports controllers.

<details><summary>Example abstraction / fix</summary>
<p>

Assuming you have already read the `API Prototype / Pseudo Code` section, we could introduce our own `InputState`:

```java
/**
 * Abstract player input state as a record.
 * Mirrors vanilla Minecraft {@link Input} class but works for any controller.
 */
@OnlyIn(Dist.CLIENT)
public record InputState(
        float leftImpulse,
        float forwardImpulse,
        boolean up,
        boolean down,
        boolean left,
        boolean right,
        boolean jumping,
        boolean sneaking
) {
    /**
     * Returns a 2D movement vector (left/right, forward/back) based on impulses.
     */
    public Vec2 getMoveVector() {
        return new Vec2(this.leftImpulse, this.forwardImpulse);
    }

    /**
     * Returns true if there is any forward movement.
     */
    public boolean hasForwardImpulse() {
        return this.forwardImpulse > 1.0E-5F;
    }
}
```

Add a get method to the interface:

```java
public interface IEpicFightController {
    boolean inControllerMode();

    boolean isActionJustPressed(EpicFightInputAction action);

    boolean isActionTriggered(EpicFightInputAction action);
    
    @Nullable
    InputState getInputState();
}
```

Implement it with Controlify mod specific-API:

```java
public class ControlifyEpicFightController implements IEpicFightController {
    private static ControlifyApi getApi() {
        return ControlifyApi.get();
    }

    // ...

    @Nullable
    @Override
    public InputState getInputState() {
        Optional<ControllerEntity> optionalControllerEntity = getApi().getCurrentController();

        if (optionalControllerEntity.isEmpty()) {
            return null;
        }

        ControllerEntity controller = optionalControllerEntity.get();

        InputBinding forwardBind = ControlifyBindings.WALK_FORWARD.on(controller);
        InputBinding backwardBind = ControlifyBindings.WALK_BACKWARD.on(controller);
        InputBinding leftBind = ControlifyBindings.WALK_LEFT.on(controller);
        InputBinding rightBind = ControlifyBindings.WALK_RIGHT.on(controller);
        InputBinding jumpBind = ControlifyBindings.JUMP.on(controller);
        InputBinding sneakBind = ControlifyBindings.SNEAK.on(controller);

        float forwardImpulse = forwardBind.analogueNow() - backwardBind.analogueNow();
        float leftImpulse = rightBind.analogueNow() - leftBind.analogueNow();

        return new InputState(leftImpulse, forwardImpulse, forwardBind.digitalNow(), backwardBind.digitalNow(), leftBind.digitalNow(), rightBind.digitalNow(), jumpBind.digitalNow(), sneakBind.digitalNow());
    }
}
```

Now we can use this in `EpicFightControls`:

```java
@OnlyIn(Dist.CLIENT)
public final class EpicFightControls {
    private static final IEpicFightController controllerInputProvider = new ControlifyEpicFightController();

   // ...

    public static InputState getInputState() {
        if (isUsingController()) {
            InputState inputState = controllerInputProvider.getInputState();

            if (inputState != null) {
                return inputState;
            }
        }

        LocalPlayer player = MC.player;
        Input playerInput = player.input;
        return new InputState(
                playerInput.leftImpulse, playerInput.forwardImpulse,
                playerInput.up, playerInput.down,
                playerInput.left, playerInput.right,
                playerInput.jumping, playerInput.shiftKeyDown
        );
    }
}
```

Now we can use `EpicFightControls.getInputState` in `DodgeSkill`, for example:

```diff
+ InputState inputState = EpicFightControls.getInputState();
+ int forward = inputState.up() ? 1 : 0;
+ int backward = inputState.down() ? -1 : 0;
+ int left = inputState.left() ? 1 : 0;
+ int right = inputState.right() ? -1 : 0;
- int forward = input.up ? 1 : 0;
- int backward = input.down ? -1 : 0;
- int left = input.left ? 1 : 0;
- int right = input.right ? -1 : 0;
```

We need to test with more controller mods to ensure the fix is universal.

</p>
</details> 

#### Violations / What Needs Fixing

We should avoid making these direct calls and instead rely on the new input system, which supports input from either the keyboard or a controller:

* `isKeyDown` in [`ControlEngine.handleEpicFightKeyMappings`](https://github.com/Epic-Fight/epicfight/blob/9cf2e7918daa64c82d8496134c4ca3a12ced1fb5/src/main/java/yesman/epicfight/client/events/engine/ControlEngine.java#L276) to fix weapon innate skill
* `Minecraft.getInstance().options.keyJump.isDown()` in [`PhantomAscentSkill.movementInputEvent`](https://github.com/Epic-Fight/epicfight/blob/9cf2e7918daa64c82d8496134c4ca3a12ced1fb5/src/main/java/yesman/epicfight/skill/mover/PhantomAscentSkill.java#L69) to fix the double jump
* `.up`, `.down`, `.left`, or `.right` from `Input` in [`PhantomAscentSkill.movementInputEvent`](https://github.com/Epic-Fight/epicfight/blob/9cf2e7918daa64c82d8496134c4ca3a12ced1fb5/src/main/java/yesman/epicfight/skill/mover/PhantomAscentSkill.java#L101-L104) to determine player movement during a double jump.
* `.up`, `.down`, `.left`, or `.right` from `Input` in [`DodgeSkill.gatherArguments`](https://github.com/Epic-Fight/epicfight/blob/9cf2e7918daa64c82d8496134c4ca3a12ced1fb5/src/main/java/yesman/epicfight/skill/dodge/DodgeSkill.java#L62-L65) to determine player movement during a dodge.
* `.left` `.right` in [`KnockdownWakeupSkill.gatherArguments`](https://github.com/Epic-Fight/epicfight/blob/9cf2e7918daa64c82d8496134c4ca3a12ced1fb5/src/main/java/yesman/epicfight/skill/dodge/KnockdownWakeupSkill.java#L28-L29) to determine player movement during a knock down wake up.
* `isKeyDown` in [`ControlEngine.inputTick`](https://github.com/Epic-Fight/epicfight/blob/9cf2e7918daa64c82d8496134c4ca3a12ced1fb5/src/main/java/yesman/epicfight/client/events/engine/ControlEngine.java#L404), part of the Demolition leap fix.
* `this.options.keyJump.getKey().getValue()` in [`ControlEngine.handleEpicFightKeyMappings`](https://github.com/Epic-Fight/epicfight/blob/9cf2e7918daa64c82d8496134c4ca3a12ced1fb5/src/main/java/yesman/epicfight/client/events/engine/ControlEngine.java#L233) to determine whether the player is holding Space and Epic Fight mover keybind is the same as the vanilla one. This is currently always `false` when using a controller mod, so Demolition Leap will never be executed.
* `playerpatch.getOriginal().input.forwardImpulse` and `playerpatch.getOriginal().input.leftImpulse` in: [`ComboAttackAnimation.shouldPlayerMove`](https://github.com/Epic-Fight/epicfight/blob/9cf2e7918daa64c82d8496134c4ca3a12ced1fb5/src/main/java/yesman/epicfight/api/animation/types/ComboAttackAnimation.java#L140) to fix a super minor issue when game rule `STIFF_COMBO_ATTACKS` is `true` (defaults to `false`).
* Apply [these fixes](https://github.com/Epic-Fight/epicfight/blob/15c66498ac05f6a717f7e69ceaacd214e9846822/src/main/java/yesman/epicfight/client/events/engine/ControlEngine.java#L537-L603) (e.g., prevent item switch during demolition leap) to controllers as well.
* There are other violations in `ControlEngine`, but the above is minimal to get the basic features working.

#### Optional / Minor

Minor improvements to better support controllers:

* **Disable controller player look inputs during target lock-on:**  
  This was fixed for the mouse in #1998, but controllers were not addressed. Implementing this may require a controller-specific mod API, and it’s unclear if it can be easily abstracted.

  <details><summary>Example fix (only for Controlify)</summary>
  <p>
  
  ```java
  @OnlyIn(Dist.CLIENT)
  @Mixin(InGameInputHandler.class)
  public class MixinInGameInputHandler {
      @Inject(method = "processPlayerLook(F)V", at = @At(value = "INVOKE", target = "Lnet/minecraft/client/player/LocalPlayer;turn(DD)V", shift = At.Shift.BEFORE), cancellable = true)
      private static void inject(float deltaTime, CallbackInfo ci) {
          LocalPlayerPatch localPlayerPatch = ClientEngine.getInstance().getPlayerPatch();
  
          if (localPlayerPatch != null && localPlayerPatch.isTargetLockedOn()) {
              // Fixes a minor issue, though we should rely on Controlify supported API if available
              ci.cancel();
          }
      }
  }
  ```

  Controlify also supports [LOOK_INPUT_MODIFIER](https://github.com/isXander/Controlify/blob/3d5c7383c4a4c07f35c598b1e3f5d890d343bb0a/src/main/java/dev/isxander/controlify/api/event/ControlifyEvents.java#L39-L43) event; we should consider using it before using mixins:

  ```java
  ControlifyEvents.LOOK_INPUT_MODIFIER.register(event -> {
              LocalPlayerPatch localPlayerPatch = ClientEngine.getInstance().getPlayerPatch();
  
              if (localPlayerPatch != null && localPlayerPatch.isTargetLockedOn()) {
                  // Fixes a minor issue when locking on an enemy.
                  event.lookInput().zero();
              }
          });
  ```
  
  </p>
  </details> 

* **[Very low priority] Cape physics don't work with some controller mods:**  
Cape physics works perfectly fine with Controlify mod, but not with [Controllable](https://www.curseforge.com/minecraft/mc-mods/controllable) mod.

  I suspect [`EntitySnapshot.PlayerSnapshot`](https://github.com/Epic-Fight/epicfight/blob/9cf2e7918daa64c82d8496134c4ca3a12ced1fb5/src/main/java/yesman/epicfight/api/utils/EntitySnapshot.java#L277-L288). The problem isn’t with rendering the cape itself—it’s fine for vanilla input. The issue is that the cape physics are being fed vanilla movement info. 

  FYI: Controlify (and other mods too) has [custom player movement implementation](https://github.com/isXander/Controlify/blob/e9a1e863548ab3a78f22e1710c69ad21c9e5bcad/src/main/java/dev/isxander/controlify/ingame/ControllerPlayerMovement.java#L19) made for controllers. This is likely the case with Controllable mod, though their custom movement info isn't compatible with Epic Fight cape physics. I'm not 100% if the issue is on their side, but this is a super minor issue and we should not focus on it right now.

* **Add a custom bind context for Epic Fight combat mode:**
  This requires a Controlify-specific API. We should register a custom bind context so that some buttons only apply in Epic Fight combat/battle mode, rather than mining. This will support contextual button prompts so they are better displayed in the GUI. This change is to align with [the existing `CombatKeyMapping`](https://github.com/Epic-Fight/epicfight/blob/9cf2e7918daa64c82d8496134c4ca3a12ced1fb5/src/main/java/yesman/epicfight/client/input/CombatKeyMapping.java#L16-L19)

* **[Very Low priority for now] Add guides:**
  This definitely requires a Controlify-specific API. [More details](https://moddedmc.wiki/en/project/controlify/latest/docs/resource-packs/guides).

<details><summary>Example</summary>
<p>
  
```java
  public static boolean isInGame(Minecraft mc) {
        return mc.screen == null && mc.level != null && mc.player != null;
    }
    public static final BindContext COMBAT_MODE = new BindContext(
            EpicFightMod.id("combat_mode"),
            mc -> isInGame(mc) && ClientEngine.getInstance().isEpicFightMode()
    );

  public static InputBindingSupplier COMBAT_ATTACK;

   COMBAT_ATTACK = ControlifyBindApi.get().registerBinding(
                builder -> builder
                        .id(EpicFightMod.MODID, "combatAttack")
                        .category(ControlifyBindings.GAMEPLAY_CATEGORY)
                        .allowedContexts(COMBAT_MODE)
                        .description(Component.literal("The attack button for Epic Fight combat mode"))
                        .name(Component.literal("Combat Mode Attack"))
                        .addKeyCorrelation(EpicFightKeyMappings.ATTACK) // Remove the vanilla Epic Fight attack key but only for controllers
        );

//  FYI: There is also this workaround that doesn't even require to update ControlEngine but it's best to avoid it for best experience and do more work instead: .keyEmulation(EpicFightKeyMappings.ATTACK) // Workaround, it's probably better to update ControlEngine
``` 

</p>
</details> 

  **This is is not a priority at the moment since the goal is to focus on getting the core done.**

* **[Low priority for now] Add support for skill book and skill tree screens:**
  This requires Controlify specific API to fix properly, though most mods support virtual cursor, but this isn't ideal for a good user experience.

#### Also?

It's likely that we need to introduce some changes in:

* `ControlEngine.makeUnpressed` and `ControlEngine.setKeyBind`
* `ControlEngine.reservedKey`
* `ControlEngine.currentHoldingKey`
* `ControlEngine.lockHotkeys`
* `ControlEngine.epicfight$interactionKeyMappingTriggered`
* `HoldableSkill.getKeyMapping`
* `EpicFightClientBoundPayloadHandler.handleSkillFeedback`
* `CustomChargeableSkill.keyMapping`

We need more testing with additional controller mods and by trying different movements to determine if any adjustments are necessary. The goal is to ensure Epic Fight works properly with controller mods with minimal issues.

<details><summary>A better alternative to disable vanilla attack</summary>
<p>

I found a more efficient way to disable the vanilla attack that works across all input systems.  

Currently, Epic Fight uses a method in `ControlEngine` to disable attacks and item swapping, but it only works for keyboard input and must be called before `Minecraft#handleKeybinds`:

```java
//  This method is called before `Minecraft#handleKeybinds` is called.
public static void makeUnpressed(KeyMapping keyMapping) {
		while (keyMapping.consumeClick()) {}
		setKeyBind(keyMapping, false);
	}
```

A better approach is to cancel the vanilla attack at its source, which works for controllers and other mods like Controllable, not just Controlify.  :

```java
@Inject(method = "startAttack", at = @At("HEAD"), cancellable = true)
private void onVanillaAttack(CallbackInfoReturnable<Boolean> cir) {
    cir.cancel();
}
```

</p>
</details> 

### Questions

<details><summary>Question 1?</summary>
<p>

I already asked about this in [this comment](https://github.com/Epic-Fight/epicfight/pull/2112#discussion_r2432547726). I need to understand whether there is a reason why the code sometimes uses `ControlEngine.isKeyDown` and other times uses the vanilla `KeyMapping.isDown`. Note that `ControlEngine.isKeyDown` calls `KeyMapping.isDown` internally.

</p>
</details> 

<details><summary>Question 2?</summary>
<p>

Why does Epic Fight registers a custom attack keybind? It looks like it was added in [this commit](https://github.com/Epic-Fight/epicfight/compare/e963d283e2ec...35e8aa9ea4ba), was this added just as a workaround to fix Controller mods compatibility?

</p>
</details> 

### Reference

Here are some references to other mods and Controlify

* [ControlifyCompat.java (lines 2-148) – Do a Barrel Roll](https://github.com/enjarai/do-a-barrel-roll/blob/stonecutter/1.21/src/main/java/nl/enjarai/doabarrelroll/compat/controlify/ControlifyCompat.java#L2-L148)  

* [Automobility – ControlifyCompat.java](https://github.com/FoundationGames/Automobility/blob/1.21-rewrite/fabric/src/main/java/io/github/foundationgames/automobility/fabric/controller/controlify/ControlifyCompat.java) and [Automobility – AutomobileController.java (lines 3-48)](https://github.com/FoundationGames/Automobility/blob/1.21-rewrite/common%2Fsrc%2Fmain%2Fjava%2Fio%2Fgithub%2Ffoundationgames%2Fautomobility%2Fcontroller%2FAutomobileController.java#L3-L48) .
* https://moddedmc.wiki/en/project/controlify/latest/docs/developers/bindings-api (slightly outdated for NeoForge but I know the differences, see https://github.com/isXander/Controlify/issues/696)

### Controller mods

See also full mod comparison: https://moddedmc.wiki/en/project/controlify/latest/docs/users/mod-comparison

These are the controller mods that we should test Epic Fight against. The goal is not to support them all, but to ensure our abstractions are well-designed and universal.

#### [Controlify](https://moddedmc.wiki/en/project/controlify/latest)

* **Availability**: The oldest supported Minecraft versions are NeoForge 1.20.3 and Fabric 1.19.3. The mod is usually ported quickly and is already available up to version 1.21.10. Though does not support Forge 1.20.1 even with Sinytra connector.
* **Public API**: Yes, has exactly what we need and works well ([Bindings API](https://moddedmc.wiki/en/project/controlify/latest/docs/developers/bindings-api)).
* **Input binding implementation**: [InputBinding](https://github.com/isXander/Controlify/blob/multiversion/dev/src/main/java/dev/isxander/controlify/api/bind/InputBinding.java#L26)

#### [Controllable](https://www.curseforge.com/minecraft/mc-mods/controllable)

* **Availability**: Supports both NeoForge 1.21.1 and Forge 1.20.1
* **Public API**: No
* **Input binding implementation**:  [ButtonBinding](https://github.com/MrCrayfish/Controllable/blob/multiloader/1.21.9/common/src/main/java/com/mrcrayfish/controllable/client/binding/ButtonBinding.java#L12)

#### [Controller Support Mod](https://www.curseforge.com/minecraft/mc-mods/controller-mod)

* **Availability**: Supports both NeoForge 1.21.1 and Forge 1.20.1
* **Public API**: No
* **Input binding implementation**:  **The code repository is private.**

#### [MidnightControls](https://modrinth.com/mod/midnightcontrols)

* **Availability**: Supports only NeoForge 1.21.1.
* **Public API**: No, though the internal API seems to be more stable than Controllable and Controller Support mod.
* **Input binding implementation**:  [ButtonBinding](https://github.com/TeamMidnightDust/MidnightControls/blob/architectury-1.21.9/common/src/main/java/eu/midnightdust/midnightcontrols/client/controller/ButtonBinding.java#L39)

See also PR #2122 for the current real proposed system

EchoEllet:
This PR fixes #2116 and aims to simplify and improve controller support; it does not target any specific controller mod at this stage. More PRs are in progress.

Feedback, suggestions, and questions are welcome at any stage.

#### Input Actions

```diff
- EpicFightKeyMappings.DODGE; // Raw access to KeyMapping, does not support controllers.
+ EpicFightInputActions.DODGE; // Use raw ".keyMapping()" if really necessary; prefer new supported APIs.
```

#### Discrete (one-time) actions

```diff
- while (isKeyPressed(EpicFightKeyMappings.ATTACK, true)) {
-    // ...
- }
+ InputManager.triggerOnPress(EpicFightInputActions.ATTACK, true, (context) -> {
+      boolean triggeredByController = context.triggeredByController() // Optional context
+     // ...
+ });
```

#### Continuous actions

```diff
- if (minecraft.options.keyAttack.isDown()) {}
+ if (InputManager.isActionActive(EpicFightInputActions.VANILLA_ATTACK_DESTROY)) {}
```

##### Direction

```diff
- int forward = input.up ? 1 : 0;
- int backward = input.down ? -1 : 0;
- int left = input.left ? 1 : 0;
- int right = input.right ? -1 : 0;
- int vertic = forward + backward;
- int horizon = left + right;
+ final MovementDirection movementDirection = MovementDirection.fromInputState(InputManager.getInputState(input));
+ final int vertic = movementDirection.vertical();
+ final int horizon = movementDirection.horizontal();
```

##### Updating state

```diff
- Minecraft.getInstance().player.input.jumping = true; // Bad: Mutable, unpredictable, and may not support controllers
+ final PlayerInputState updated = InputManager.getInputState(Minecraft.getInstance().player)
+        .withJumping(true);
+ InputManager.setInputState(updated);
```

#### Checking if bound to the same physical button/key

```diff
- if (EpicFightKeyMappings.WEAPON_INNATE_SKILL.getKey().equals(EpicFightKeyMappings.ATTACK.getKey())) {}
+ if (InputManager.isBoundToSamePhysicalInput(EpicFightInputActions.WEAPON_INNATE_SKILL, EpicFightInputActions.ATTACK)) {}
```

#### Integrating a controller mod

```java
// Controlify mod is used as an example here.
final class ControlifyModIntegration implements IEpicFightControllerMod {
    // Implement the required methods...
}
```

Then register it:

```java
EpicFightControllerModProvider.set(EpicFightMod.MODID, new ControlifyModIntegration());
```

#### Low-level Controller access

```java
IEpicFightControllerMod controllerModApi = EpicFightControllerModProvider.get();

if (controllerModApi == null {
    // Null if no controller mod integration is available
   // (e.g., the controller mod is not installed, or Epic Fight does not support it).
    return;
}

// Checking the input mode

InputMode inputMode = controllerModApi.getInputMode();
boolean supportsController = inputMode.supportsController();

// Getting raw controller binding

ControllerBinding moveForward = controllerModApi.getBinding(EpicFightInputActions.MOVE_FORWARD);
float analogue = moveForward.getAnalogueNow();

// OR

ControllerBinding jump = controllerModApi.getBinding(EpicFightInputActions.JUMP);
boolean justPressed = jump.isDigitalJustPressed();

```

### Additional Cleanup Checklist

- [x] `ControlEngine.handleEpicFightKeyMappings` currently mixes **when** an action should be triggered with **how** it is executed. To simplify future improvements and integration with the new input system, the **when** logic—including the conditional checks and associated `while`/`if` statements—should be extracted into separate functions.  
- [x] `ControlEngine.handleEpicFightKeyMappings` is long and not modular. Consider breaking it into smaller, focused methods (e.g., `handleAttackKey()`, `handleDodgeKey()`, `handleGuardKey()`) to improve readability, and maintainability.

### Very Minor issues

These are existing issues when using a controller mod with Epic Fight. These should be tested and fixed with Controlify on NeoForge 1.21.1:

- [x] Player should not be able to switch or drop the current selected item when Epic Fight locks the hotbar.
- [ ] Should not be able to place Epic Fight greatswords or any incompatible weapon types in the offhand slot within the inventory GUI.

These issues don't affect the player experience significantly. Most players will not notice them.

### Changelog

We should add a note to the changelog of the new version if it includes these changes:

```md
**Important:** If you're a mod developer modifying Epic Fight’s `ControlEngine` using a mixin, please test your changes against the latest version to avoid breaking Epic Fight internals or your add-on’s behavior.
You may see some **deprecations** — please consider **migrating** to support controllers in future updates.
For more details, refer to [PR #2122](https://github.com/Epic-Fight/epicfight/pull/2122).
```

### Testing

Verified using manual testing against many Epic Fight addons:

https://github.com/user-attachments/assets/cb24f9a1-74d6-43ef-9286-33256cbaf2bd

### DIFF
--- a/src/main/java/yesman/epicfight/api/client/input/PlayerInputState.java
+++ b/src/main/java/yesman/epicfight/api/client/input/PlayerInputState.java
@@ -1,0 +1,113 @@
+package yesman.epicfight.api.client.input;
+
+import net.minecraft.client.player.Input;
+import net.minecraft.world.phys.Vec2;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Represents the abstract state of the player input (client-side).
+ * <p>
+ * Mirrors the vanilla Minecraft {@link Input} class.
+ * </p>
+ */
+public record PlayerInputState(
+        float leftImpulse,
+        float forwardImpulse,
+        boolean up,
+        boolean down,
+        boolean left,
+        boolean right,
+        boolean jumping,
+        boolean sneaking
+) {
+    public static PlayerInputState fromVanillaInput(Input input) {
+        return new PlayerInputState(
+                input.leftImpulse, input.forwardImpulse,
+                input.up, input.down,
+                input.left, input.right,
+                input.jumping, input.shiftKeyDown
+        );
+    }
+
+    /**
+     * Applies the values from a {@link PlayerInputState} to a vanilla {@link Input} instance.
+     * **Note:** Updating the vanilla {@link Input} fields has side effects, so this change is **not immutable**.
+     *
+     * @param updated the new input state to apply
+     * @param input   the existing vanilla Input instance to update
+     */
+    public static Input applyToVanillaInput(@NotNull PlayerInputState updated, @NotNull Input input) {
+        if (input.leftImpulse != updated.leftImpulse()) input.leftImpulse = updated.leftImpulse();
+        if (input.forwardImpulse != updated.forwardImpulse()) input.forwardImpulse = updated.forwardImpulse();
+        if (input.up != updated.up()) input.up = updated.up();
+        if (input.down != updated.down()) input.down = updated.down();
+        if (input.left != updated.left()) input.left = updated.left();
+        if (input.right != updated.right()) input.right = updated.right();
+        if (input.jumping != updated.jumping()) input.jumping = updated.jumping();
+        if (input.shiftKeyDown != updated.sneaking()) input.shiftKeyDown = updated.sneaking();
+        return input;
+    }
+
+    public Vec2 getMoveVector() {
+        return new Vec2(this.leftImpulse, this.forwardImpulse);
+    }
+
+    public boolean hasForwardImpulse() {
+        return this.forwardImpulse > 1.0E-5F;
+    }
+
+    public @NotNull PlayerInputState copyWith(
+            @Nullable Float leftImpulse,
+            @Nullable Float forwardImpulse,
+            @Nullable Boolean up,
+            @Nullable Boolean down,
+            @Nullable Boolean left,
+            @Nullable Boolean right,
+            @Nullable Boolean jumping,
+            @Nullable Boolean sneaking
+    ) {
+        return new PlayerInputState(
+                leftImpulse != null ? leftImpulse : this.leftImpulse,
+                forwardImpulse != null ? forwardImpulse : this.forwardImpulse,
+                up != null ? up : this.up,
+                down != null ? down : this.down,
+                left != null ? left : this.left,
+                right != null ? right : this.right,
+                jumping != null ? jumping : this.jumping,
+                sneaking != null ? sneaking : this.sneaking
+        );
+    }
+
+    public @NotNull PlayerInputState withLeftImpulse(float leftImpulse) {
+        return copyWith(leftImpulse, null, null, null, null, null, null, null);
+    }
+
+    public @NotNull PlayerInputState withForwardImpulse(float forwardImpulse) {
+        return copyWith(null, forwardImpulse, null, null, null, null, null, null);
+    }
+
+    public @NotNull PlayerInputState withUp(boolean up) {
+        return copyWith(null, null, up, null, null, null, null, null);
+    }
+
+    public @NotNull PlayerInputState withDown(boolean down) {
+        return copyWith(null, null, null, down, null, null, null, null);
+    }
+
+    public @NotNull PlayerInputState withLeft(boolean left) {
+        return copyWith(null, null, null, null, left, null, null, null);
+    }
+
+    public @NotNull PlayerInputState withRight(boolean right) {
+        return copyWith(null, null, null, null, null, right, null, null);
+    }
+
+    public @NotNull PlayerInputState withJumping(boolean jumping) {
+        return copyWith(null, null, null, null, null, null, jumping, null);
+    }
+
+    public @NotNull PlayerInputState withSneaking(boolean sneaking) {
+        return copyWith(null, null, null, null, null, null, null, sneaking);
+    }
+}
--- a/src/main/java/yesman/epicfight/api/client/input/handlers/InputManager.java
+++ b/src/main/java/yesman/epicfight/api/client/input/handlers/InputManager.java
@@ -1,0 +1,225 @@
+package yesman.epicfight.api.client.input.handlers;
+
+import com.mojang.blaze3d.platform.InputConstants;
+import net.minecraft.client.KeyMapping;
+import net.minecraft.client.Minecraft;
+import net.minecraft.client.player.Input;
+import net.minecraft.client.player.LocalPlayer;
+import net.minecraft.world.InteractionHand;
+import net.minecraftforge.client.ForgeHooksClient;
+import net.minecraftforge.client.event.InputEvent;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.lwjgl.glfw.GLFW;
+import yesman.epicfight.api.client.input.InputMode;
+import yesman.epicfight.api.client.input.PlayerInputState;
+import yesman.epicfight.api.client.input.action.EpicFightInputActions;
+import yesman.epicfight.api.client.input.controller.EpicFightControllerModProvider;
+import yesman.epicfight.api.client.input.controller.IEpicFightControllerMod;
+import yesman.epicfight.client.events.engine.ControlEngine;
+import yesman.epicfight.api.client.input.controller.ControllerBinding.InputType;
+
+/**
+ * High-level input API that abstracts direct interactions with {@link KeyMapping}
+ * and supports controllers if an Epic Fight controller mod implementation is present
+ * (see {@link EpicFightControllerModProvider}).
+ * <p>
+ * Use this class whenever possible to ensure input works consistently across
+ * keyboard/mouse and supported controllers.
+ */
+@ApiStatus.Experimental
+public final class InputManager {
+    private InputManager() {
+    }
+
+    @Nullable
+    private static IEpicFightControllerMod getControllerModApi() {
+        return EpicFightControllerModProvider.get();
+    }
+
+    /**
+     * Checks if controller or gamepad input is currently supported.
+     *
+     * <p><b>Note:</b> The {@link InputMode#MIXED} mode supports both controller and keyboard/mouse input at the same time.
+     * Returning {@code true} here does not necessarily mean the input mode is exclusively {@link InputMode#CONTROLLER}.
+     *
+     * @return {@code true} if a controller mod is present and the current input mode allows controller input.
+     * @see InputMode
+     */
+    public static boolean supportsControllerInput() {
+        final IEpicFightControllerMod controllerMod = getControllerModApi();
+        if (controllerMod == null) {
+            return false;
+        }
+        return controllerMod.getInputMode().supportsController();
+    }
+
+    /**
+     * Returns whether the given input action is currently active this tick.
+     * <p>
+     * Checks the action state depending on the current input mode:
+     * <ul>
+     *     <li>{@link InputMode#KEYBOARD_MOUSE}: whether the key mapping is down.</li>
+     *     <li>{@link InputMode#CONTROLLER}: whether the controller binding digital state is active.</li>
+     *     <li>{@link InputMode#MIXED}: true if either the key mapping is down or the controller binding state is active.</li>
+     * </ul>
+     * If no controller mod is present, only the key mapping is checked.
+     * This is usually useful for continuous actions.
+     *
+     * @param action the input action to check
+     * @return true if the action is currently this tick active; false otherwise
+     * @see InputType
+     */
+    public static boolean isActionActive(@NotNull EpicFightInputActions action) {
+        final IEpicFightControllerMod controllerMod = getControllerModApi();
+        if (controllerMod == null) {
+            return isKeyDown(action.keyMapping());
+        }
+
+        return switch (controllerMod.getInputMode()) {
+            case KEYBOARD_MOUSE -> isKeyDown(action.keyMapping());
+            case CONTROLLER -> controllerMod.getBinding(action).isDigitalActiveNow();
+            case MIXED -> isKeyDown(action.keyMapping()) || controllerMod.getBinding(action).isDigitalActiveNow();
+        };
+    }
+
+    /**
+     * Called on every client tick to potentially trigger the provided callback for a given input action.
+     * <p>
+     * This calls the internal {@link DiscreteInputActionTrigger#triggerOnPress} API, and additionally fires
+     * the {@link InputEvent.InteractionKeyMappingTriggered} input event for keyboard/mouse if {@code interactionKeyEventCheck} is {@code true}.
+     *
+     * @param action     The input action to monitor and trigger.
+     * @param interactionKeyEventCheck If {@code true}, fires the {@link InputEvent.InteractionKeyMappingTriggered} event for non-controller actions.
+     *                   This event is cancellable.
+     * @param handler    The callback to invoke when the action triggers.
+     * @see DiscreteInputActionTrigger#triggerOnPress Internal implementation details.
+     */
+    public static void triggerOnPress(@NotNull EpicFightInputActions action, boolean interactionKeyEventCheck, @NotNull DiscreteActionHandler handler) {
+        DiscreteInputActionTrigger.triggerOnPress(action, (context) -> {
+            if (context.triggeredByController() || !interactionKeyEventCheck) {
+                handler.onAction(context);
+                return;
+            }
+
+            runKeyboardMouseEvent(action, handler);
+        });
+    }
+
+    /**
+     * Convenience overload of {@link #triggerOnPress(EpicFightInputActions, boolean, DiscreteActionHandler)}
+     * for callbacks that do not require the {@link DiscreteActionHandler.Context}.
+     *
+     * @see #triggerOnPress(EpicFightInputActions, boolean, DiscreteActionHandler)
+     */
+    public static void triggerOnPress(@NotNull EpicFightInputActions action, boolean interactionKeyEventCheck, @NotNull Runnable runnable) {
+        triggerOnPress(action, interactionKeyEventCheck, (context) -> runnable.run());
+    }
+
+    /**
+     * Checks whether the given input action is assigned to the same key / button as another action.
+     * <p>
+     * For keyboard/mouse, this compares the key codes; for controllers, it compares the digital button.
+     * <p><b>Note:</b> {@link InputMode#MIXED} is currently unsupported and its behavior is undefined.</p>
+     *
+     * @param action  the first input action
+     * @param action2 the second input action
+     * @return true if both actions are triggered by the same key or controller button; false otherwise
+     */
+    public static boolean isBoundToSamePhysicalInput(@NotNull EpicFightInputActions action, @NotNull EpicFightInputActions action2) {
+        final IEpicFightControllerMod controllerMod = getControllerModApi();
+        if (controllerMod != null && controllerMod.getInputMode() == InputMode.CONTROLLER) {
+            return controllerMod.isBoundToSameButton(action, action2);
+        }
+
+        final KeyMapping keyMapping1 = action.keyMapping();
+        final KeyMapping keyMapping2 = action2.keyMapping();
+        return keyMapping1.getKey() == keyMapping2.getKey();
+    }
+
+    /**
+     * Retrieves the current input state for the current player (client-side).
+     * <p><b>Note:</b> {@link InputMode#MIXED} is currently unsupported and its behavior is undefined.</p>
+     * You should use this method instead of depending on the vanilla {@link Input} directly
+     * to support controllers.
+     * <p>
+     * The {@link PlayerInputState} is immutable, so properties cannot be updated directly, for that,
+     * use {@link InputManager#setInputState}.
+     *
+     * @param vanillaInput the Minecraft vanilla {@link Input} which will be mapped to a {@link PlayerInputState};
+     *                     ignored if using a controller.
+     * @return an immutable {@link PlayerInputState} representing the current input state.
+     * @see InputManager#setInputState
+     */
+    public static PlayerInputState getInputState(@NotNull Input vanillaInput) {
+        final IEpicFightControllerMod controllerMod = getControllerModApi();
+        if (controllerMod != null && controllerMod.getInputMode() == InputMode.CONTROLLER) {
+            return controllerMod.getInputState();
+        }
+
+        return PlayerInputState.fromVanillaInput(vanillaInput);
+    }
+
+    /**
+     * Updates the current input state for the current player (client-side).
+     * <p>
+     * You should use this instead of modifying fields in the vanilla {@link Input} directly
+     * to ensure controller input is properly supported.
+     * </p>
+     *
+     * @param inputState the updated input state.
+     * @see InputManager#getInputState
+     */
+    public static void setInputState(@NotNull PlayerInputState inputState) {
+        final LocalPlayer player = Minecraft.getInstance().player;
+        if (player != null) {
+            Input input = player.input;
+            PlayerInputState.applyToVanillaInput(inputState, input);
+        }
+    }
+
+    /**
+     * Handles firing the {@link InputEvent.InteractionKeyMappingTriggered} input event for keyboard/mouse actions
+     * and runs the callback only if the event is not canceled.
+     * This method replaces the legacy internal {@link ControlEngine#isKeyPressed}.
+     */
+    @SuppressWarnings("JavadocReference")
+    private static void runKeyboardMouseEvent(@NotNull EpicFightInputActions action, @NotNull DiscreteActionHandler handler) {
+        final KeyMapping keyMapping = action.keyMapping();
+
+        final InputConstants.Key key = keyMapping.getKey();
+        final boolean isMouse = InputConstants.Type.MOUSE == key.getType();
+
+        final int mouseButton = isMouse ? key.getValue() : -1;
+
+        @SuppressWarnings("UnstableApiUsage")
+        InputEvent.InteractionKeyMappingTriggered inputEvent = ForgeHooksClient.onClickInput(
+                mouseButton, keyMapping, InteractionHand.MAIN_HAND
+        );
+
+        if (!inputEvent.isCanceled()) {
+            handler.onAction(new DiscreteActionHandler.Context(false));
+        }
+    }
+
+    /**
+     * Checks whether the vanilla {@link KeyMapping} is down.
+     * This internal method replaces the legacy {@link ControlEngine#isKeyDown}.
+     */
+    private static boolean isKeyDown(@NotNull KeyMapping keyMapping) {
+        if (keyMapping.isDown()) {
+            return true;
+        }
+        final InputConstants.Key key = keyMapping.getKey();
+        final int keyValue = key.getValue();
+        final long windowPointer = Minecraft.getInstance().getWindow().getWindow();
+
+        if (key.getType() == InputConstants.Type.KEYSYM) {
+            return GLFW.glfwGetKey(windowPointer, keyValue) > 0;
+        } else if (key.getType() == InputConstants.Type.MOUSE) {
+            return GLFW.glfwGetMouseButton(windowPointer, keyValue) > 0;
+        }
+        return false;
+    }
+}
--- a/src/main/java/yesman/epicfight/api/client/input/handlers/InputManager.java
+++ b/src/main/java/yesman/epicfight/api/client/input/handlers/InputManager.java
@@ -68,7 +68,7 @@ public final class InputManager {
      * This is usually useful for continuous actions.
      *
      * @param action the input action to check
-     * @return true if the action is currently this tick active; false otherwise
+     * @return true if the action is currently active in this tick; false otherwise
      * @see InputType
      */
     public static boolean isActionActive(@NotNull EpicFightInputActions action) {
@@ -152,6 +152,7 @@ public final class InputManager {
      * @return an immutable {@link PlayerInputState} representing the current input state.
      * @see InputManager#setInputState
      */
+    @NotNull
     public static PlayerInputState getInputState(@NotNull Input vanillaInput) {
         final IEpicFightControllerMod controllerMod = getControllerModApi();
         if (controllerMod != null && controllerMod.getInputMode() == InputMode.CONTROLLER) {
@@ -159,6 +160,18 @@ public final class InputManager {
         }
 
         return PlayerInputState.fromVanillaInput(vanillaInput);
+    }
+
+    /**
+     * Convenience overload of {@link #getInputState(Input)} that requires the full {@link LocalPlayer},
+     * which is needed to read the vanilla {@link Input} used for non-controller inputs.
+     *
+     * @param localPlayer the player whose vanilla {@link Input} will be read; ignored when using a controller.
+     * @return an immutable {@link PlayerInputState} representing the current input state.
+     */
+    @NotNull
+    public static PlayerInputState getInputState(@NotNull LocalPlayer localPlayer) {
+        return getInputState(localPlayer.input);
     }
 
     /**
--- a/src/main/java/yesman/epicfight/api/animation/types/BasicAttackAnimation.java
+++ b/src/main/java/yesman/epicfight/api/animation/types/BasicAttackAnimation.java
@@ -7,6 +7,8 @@ import javax.annotation.Nullable;
 
 import net.minecraft.world.InteractionHand;
 import net.minecraft.world.phys.Vec3;
+import net.minecraftforge.api.distmarker.Dist;
+import net.minecraftforge.api.distmarker.OnlyIn;
 import yesman.epicfight.api.animation.AnimationManager.AnimationAccessor;
 import yesman.epicfight.api.animation.Joint;
 import yesman.epicfight.api.animation.property.AnimationProperty.ActionAnimationProperty;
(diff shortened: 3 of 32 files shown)


EchoEllet:
> [!NOTE]
> @Yesssssman, @Forixaim, @MetalKnight56  
> If this issue is **irrelevant** to you, please add a **👎 reaction** to this comment **as an indicator**, so I won’t continue tagging you. Otherwise, please keep reading.

To make the review process smoother, I suggest splitting the work into **three stages**:

1. **Current stage:**  
   Design the public API and abstractions for the new input action system.  
   All changes are additions and **don’t modify any existing Epic Fight files**.

2. **The Hard part:**  
   Integrate the input action system into existing Epic Fight classes, mainly in `ControlEngine`, but also in other classes, such as:  
   - [`PhantomAscentSkill.java#L69`](https://github.com/Epic-Fight/epicfight/blob/9cf2e7918daa64c82d8496134c4ca3a12ced1fb5/src/main/java/yesman/epicfight/skill/mover/PhantomAscentSkill.java#L69)  
   - [`HoldableSkill.java#L47-L52`](https://github.com/Epic-Fight/epicfight/blob/9cf2e7918daa64c82d8496134c4ca3a12ced1fb5/src/main/java/yesman/epicfight/skill/modules/HoldableSkill.java#L47-L52)  
   (Documented in issue #2116)

3. **The Easy part:**  
   Integrate **Controlify** — define default controller buttons and disable player look when locking onto a target.  
   *(Controlify is only available on NeoForge and Fabric.)*
   Which means we won't be able to test the Controlify 1.21.1 Compatibility. Controllable (Forge 1.20.1) mod have issues from their side and will require mixin injections into their internals to fix. These issues don't exist in Controlify or MidnightControls (neither of which supports Forge 1.20.1).

Some of the stage 2 changes are trivial, for example:

```diff
- while (isKeyPressed(EpicFightKeyMappings.ATTACK, true)) {
-    // ...
- }
+ InputManager.triggerOnPress(EpicFightInputActions.ATTACK, true, (context) -> {
+      context.triggeredByController()
+     // ...
+ });
```

OR

```diff
- int forward = input.up ? 1 : 0;
- int backward = input.down ? -1 : 0;
- int left = input.left ? 1 : 0;
- int right = input.right ? -1 : 0;
- int vertic = forward + backward;
- int horizon = left + right;
+ final MovementDirection movementDirection = MovementDirection.fromInputState(InputManager.getInputState(input));
+ final int vertic = movementDirection.vertical();
+ final int horizon = movementDirection.horizontal();
```

Stage 1 is mostly complete. I might make improvements, small additions or fixes, but the **core design is getting there.**
Please feel free to **review** and share any **suggestions or thoughts** at this stage. I will continue developing as soon as you approve the new files.

src/main/java/yesman/epicfight/api/client/input/handlers/InputManager.java
```diff
@@ -0,0 +1,176 @@
...
+        long windowPointer = Minecraft.getInstance().getWindow().getWindow();
+
+        if (key.getType() == InputConstants.Type.KEYSYM) {
+            return GLFW.glfwGetKey(windowPointer, key.getValue()) > 0;
+        }
+        if (key.getType() == InputConstants.Type.MOUSE) {
+            return GLFW.glfwGetMouseButton(windowPointer, key.getValue()) > 0;
+        }
+        return false;
+    }
```
EchoEllet, line 237: I'm still not sure about this; sometimes Epic Fight calls [`ControlEngine#isKeyDown`](https://github.com/Epic-Fight/epicfight/blob/b2691b917a27fabe827ecf2968ff52a11319a490/src/main/java/yesman/epicfight/client/events/engine/ControlEngine.java#L479-L487) (which calls `KeyMapping.isDown` internally) and sometimes directly `KeyMapping.isDown`.

Should this new system follow that here as well? Or just always calls `ControlEngine.isKeyDown` for all key mapping checks (**current**). Both seems to work fine, so why the low-level GLFW calls?

src/main/java/yesman/epicfight/api/client/input/PlayerInputState.java
```diff
@@ -0,0 +1,112 @@
...
+public record PlayerInputState(
+        float leftImpulse,
+        float forwardImpulse,
+        boolean up,
+        boolean down,
+        boolean left,
+        boolean right,
+        boolean jumping,
+        boolean sneaking
+) {
```
EchoEllet, line 23: Even though this method seems like a lot of boilerplate code, it's actually very useful, Mojang has already introduced major breaking changes to their `Input`, so we should not depend on their `Input` across the codebase, this abstraction is useful so we can update it from exactly one place when porting to newer Minecraft versions.

So this is a useful change, not just for supporting controllers!

This initial implementation requires some extra work, but it pays off with much easier and cheaper maintenance in the long run.

src/main/java/yesman/epicfight/api/client/input/handlers/InputManager.java
EchoEllet: I'm going to solve this as solved, since I can't really reproduce any issues with both, this new method checks both the vanilla and GLFW (exactly what `ControlEngine.isKeyDown` does), so it's very unlikely that we will introduce any regression.

EchoEllet:
Actaully, let me fix these 2 super minor issues first, I don't like TODOs! I will be done soon.

EchoEllet:
@Yesssssman  Please don't merge yet into `1.20.1`, I'm almost done fixing the other super minor controller issue. I will send a PR soon.

PR #2127, opened 2025-10-23
fix: prevent player from dropping their weapon item while attacking when using a controller

EchoEllet:
- Prevents the player from accidentally dropping their weapon item while attacking when using a controller (OR any input system).
- Fixes inconsistency in my recent changes regarding the consume key clicks.
- Fixes confusing docs.

Super minor fixes, but still useful.


EchoEllet:
@Yesssssman 

Please merge this before merging or squashing `1.20.1-input-system-review`.

Yesssssman:
Checked, thanks.

PR #2128, opened 2025-10-23
Remove an outdated TODO + Minor cleanup

EchoEllet:
This does not fix any gameplay issues, just discovered that there is an outdated TODO that I already fixed, and also renamed the function `consumeHotbarSlotKeyClicks` to `disableHotbarSlotPresses`, to avoid confusion, since this one is special and does not follow the same pattern as the others, it actually do something that is keyboard-specific, unlike `consumeDropKeyClicks`, for example, which only resets the vanilla internal counter state.

This allows others and future me to read the code without any confusion. Optionally, it removes code duplication by calling an existing method.


EchoEllet:
@Yesssssman This is another PR, does not fix any bugs nor controller compatibility issues, just for maintainability reasons.
My apologies for any inconvenience.

EchoEllet:
With this PR merged, now we have only one remaining TODO that was added in #2122

PR #2141, opened 2025-10-28
[NeoForge 1.21.1] Fix Dedicated Server Crash

EchoEllet:
Similar to #2140, but for NeoForge 1.21.1.

### DIFF
--- a/src/main/java/yesman/epicfight/api/client/neoevent/MappedMovementInputUpdateEvent.java
+++ b/src/main/java/yesman/epicfight/api/client/neoevent/MappedMovementInputUpdateEvent.java
@@ -5,15 +5,17 @@ import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import net.minecraft.client.player.Input;
-import net.neoforged.api.distmarker.Dist;
-import net.neoforged.api.distmarker.OnlyIn;
 import yesman.epicfight.api.client.input.PlayerInputState;
 import yesman.epicfight.api.client.input.handlers.InputManager;
 import yesman.epicfight.api.client.input.utils.InputUtils;
 import yesman.epicfight.api.neoevent.playerpatch.PlayerPatchEvent;
 import yesman.epicfight.client.world.capabilites.entitypatch.player.LocalPlayerPatch;
 
-@OnlyIn(Dist.CLIENT)
+/**
+ * Client-side only.
+ * <p>
+ * MUST NOT be called on a dedicated server — doing so will crash the server.
+ */
 public class MappedMovementInputUpdateEvent extends PlayerPatchEvent<LocalPlayerPatch> {
 	/**
      * <b>DEPRECATED:</b>: This field is kept for backward compatibility with existing Epic Fight addons.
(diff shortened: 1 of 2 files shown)


EchoEllet:
The issue is [these lines](https://github.com/Epic-Fight/epicfight/blob/ae4afb5eabe80e18f514fbabd04aaf78f2cab167/src/main/java/yesman/epicfight/skill/SkillBuilder.java#L155-L204). They are causing `MappedMovementInputUpdateEvent` to be called or loaded somehow on the dedicated server.

EchoEllet:
I have updated the code to not depend on NeoForge's `OnlyIn` annotation for the following reasons (see also #2146):

1. NeoForge Java docs says `This is generally meant for internal Forge and FML use only and modders should avoid its use whenever possible.`. So we probably shouldn't use `OnlyIn` annotation anywhere, its only functionality is to cause a runtime crash when this class is called or inspected on the incorrect dist. [Forge also recommends not using it directly.](https://docs.minecraftforge.net/en/latest/concepts/sides/#fmlenvironmentdist-and-onlyin)
2. There is no other reliable way to fix, since in Epic Fight 1.21.1, `SkillEvent` has been introduced, which depends heavily on reflections, as soon as [getMethods()](https://github.com/Epic-Fight/epicfight/blob/ae4afb5eabe80e18f514fbabd04aaf78f2cab167/src/main/java/yesman/epicfight/skill/SkillBuilder.java#L155) is called, this crash will occur, since some usages of `SkillEvent` includes `MappedMovementInputUpdateEvent` as a parameter in the method ([example](https://github.com/Epic-Fight/epicfight/blob/ae4afb5eabe80e18f514fbabd04aaf78f2cab167/src/main/java/yesman/epicfight/skill/mover/DemolitionLeapSkill.java#L50-L51)). And while `getMethods` does not create a new instance of `MappedMovementInputUpdateEvent` (which was `OnlyIn(Dist.CLIENT)`), it inspects the class via reflections (this is a non-travial process and should be avoided for better performance and faster loading), which causes NeoForge to crash.
3. Depending on NeoForge or any dependency internals is inconvenient, as they will introduce breaking changes to code structure or even behavior. Using this annotation will just make it harder to port to other mod loaders or even newer Minecraft versions.


However:

* I have added Java docs to make it very clear that it is a client-only event; also, calling `MappedMovementInputUpdateEvent.movementInput` (vanilla`Input`) will still cause the game to crash.

I highly recommend avoiding:

* Depending on NeoForge internals directly, unless really necessary. `OnlyIn` annotation is an internal API; it might be worth filing an issue to track the process of removing these annotations everywhere
* Avoid reflections and dynamic method loading at all costs, as it causes game performance, tree-shaking, and many other runtime issues. In fact, Fabric uses fewer reflections than Forge/NeoForge, which is one reason why it loads and performs better.

EchoEllet:
@Forixaim 

Just to double check, please let me know if there is a reason why the existing code uses `OnlyIn`. According to Forge/NeoForge, Minecraft server JAR files don't contain vanilla client classes, so this is a safety check to produce a better error.

But all modded classes are already available on the server since mods usually don't ship a separate JAR file for clients & servers, so this check might not be very useful.

The game will still crash if we call a vanilla client only class/method on a dedicated server.

If you agree with my explanation, maybe we should remove `OnlyIn` annotations everywhere to be less dependent on mod loaders, and therefore make it easier to port to different mod loaders or Minecraft versions.

Maybe someday NeoForge will be forked, too. It's hard to assume or predict.

Forixaim:
I believe that OnlyIn is primarily meant to either mark something but that is primarily Yesman's coding style

EchoEllet:
Maybe we could create our own annotation that does nothing (just for the style and docs) or Javadocs for each element. But it's usually already clear since these classes are under the`client` package. 

It might be worth double checking that with @Yesssssman

Forixaim:
I personally do agree.

PR #2142, opened 2025-10-29
Merging AirAttack into BasicAttack streamlining the combat without having to use two skills to handle attacks.

Forixaim:
Changes:

- Merged AirAttack into BasicAttack and refactored ControlEngine accordingly.
- BasicAttack now consumes stamina depending on the attack. The system is rudimentary, with fixed stamina consumption for basic, dash, and air attacks.
- Chiseled Bookshelves can now accept Skillbooks
- 16x16 texture for Bokken created.


EchoEllet (commented):
Consider using Azurelib from Modrinth Maven instead of adding a JAR file to Version Control:

```groovy
compileOnly "maven.modrinth:azurelib:rkeyncYZ" // 3.1.1
```

Or even better, use [AzureLib maven](https://moddedmc.wiki/en/project/azurelib/latest/docs) to use the sources JAR  and support any transitive dependencies during development.

Forixaim:
Oh i forgot to remove it

src/main/java/yesman/epicfight/main/EpicFightMod.java
```diff
@@ -276,7 +277,14 @@ public EpicFightMod(FMLJavaModLoadingContext context) {
 		}
 		
 		if (ModList.get().isLoaded("azurelib")) {
-			ICompatModule.loadCompatModule(context, AzureLibCompat.class);
+            String test = ModList.get().getModFileById("azurelib").versionString();
+            int dotIndex = test.indexOf(".");
+            int majorVersion = Integer.parseInt(test.substring(0, dotIndex));
+            LogUtils.getLogger().debug(test);
+            if (majorVersion < 3)
+                ICompatModule.loadCompatModule(context, AzureLibCompat.class);
```
EchoEllet, line 285: It might be worth moving the version parse & compare into another utility class. It's a common use case to compare versions, also it makes sure that the `EpicFightMod` class becomes smaller and easier to read.

Forixaim: I do agree, so yes we can move on to using a util class for it.

EchoEllet: @Forixaim @Yesssssman  This has been merged. We should file an issue to clean this up or add a TODO at least to this comment, so we can clean up the mess later easily.
It would be harder if we leave this later if we don't clean OR add a TODO now.

PR #2148, opened 2025-10-30
[NeoForge 1.21.1] Implement compatibility with MCreator mods' player animations

NerdyPuzzle:
Hello, I am the creator of the MCreator plugin [Nerdy's Player Animator](https://mcreator.net/plugin/91396/player-animator-forge-1192), and in this PR I would like to introduce compatibility between MCreator mods using this plugin and epic fight as per the request of one of its users.

The animation system works between mods by having a singular MCreator mod act as the "API" which executes all the animation logic. To achieve this I have all mods using the plugin share the same string stored in the player's persistent data.

To load the compatibility layer, I check all mods' resources to find if any of the mods contain the specific file path in which the plugin stores the animation files.

Here is a video showcasing a mod which triggers player animations using the plugin working with epic fight after applying this compatibility layer:

https://github.com/user-attachments/assets/bba33917-6dc4-4957-892a-3ec325176c84

And [here](https://github.com/NerdyPuzzle/MCreator-Player-Animator/tree/main) is the link to the github repository of the MCreator plugin


Forixaim:
Hi Nerdy,
I'll test this when I have the time and get back to you after I'm done.

Sincerely,
Fori

Yesssssman:
Is it possible to optimize the checking process for the presence of the mod? So it would be like

```
if (ModList.get().getModFiles().stream().anyMatch(modFile -> {
                try {
                    Path dataPath = modFile.getFile().findResource("data");
                    return Files.exists(dataPath) && Files.list(dataPath).anyMatch(namespace -> Files.exists(namespace.resolve("bedrock_animations")));
                } catch (Exception e) {
                    return false;
                }
            }))
```

Instead of using this,

```
if (ModList.get().isLoaded("mcreatoranimations")) {
...
}
```

Yesssssman:
oh nvm since it's plugin there's no signature I can distinguish in compile time

NerdyPuzzle:
That is not possible as the animation logic is not part of any dependencies but instead a part of the mods that use the plugin. The alternative would be always loading the compatibility layer, as the check for animations playing can be done without any specific mod being loaded.

NerdyPuzzle:
> oh nvm since it's plugin there's no signature I can distinguish in compile time

Indeed, while it does act identically to how an API would by using only one of the mods to execute any logic, it is impossible to distinguish which exact mod is doing that.

PR #2155, opened 2025-11-03
[NeoForge 1.21.1] Play click sound when learn/equip a skill via controller

EchoEllet:
* When the player presses the physical controller's south button (e.g., X on a DualSense), the click sound does not play. However, when clicking the GUI button, the sound plays. This fixes the behavior consistently.
* Make `learnButton` private and expose it via a getter method for consistency with #2152, preventing consumers/mods from setting it to `null`, which could cause a crash. This is better for maintainability and long-term stability.


EchoEllet:
Also sent a PR for Epic Skills https://github.com/Epic-Fight/epicskills/pull/9

PR #2209, opened 2025-11-14
[1.21.1] Arrow keys navigation for skill editor screen

### LINKED ISSUE #2203: Enhance Skill Editor Screen Key Arrow Navigation
### History

Starting with version 1.19.4, Mojang introduced arrow key navigation, enabling seamless movement across all screens using only arrow keys (`↑`, `↓`, `→`, `←`).

The Epic Fight Skill Editor Screen was originally designed for older Minecraft versions and later ported without implementing explicit arrow navigation support.

### Issue Description

Image

To reproduce these issues, open the skill editor in-game (press `K`) and try to use the arrow keys to navigate without a mouse.

There are 4 primary issues:

1. After selecting a skill slot, it is not possible to navigate to the skill container and select a skill using the `->` key.  
2. When moving to another skill slot, the focus should switch automatically without requiring the "Enter" key. Shown in the image above.
3. Navigating through the skills should be seamless; however, it's not possible to reproduce this issue without fixing issue 1 first.
4. **[Minor]:** When selecting a skill slot via mouse left click, the focus will be changed correctly. However, when changing the current focus via the arrow key after that and then pressing "enter", the focus will be on two skill slot entries instead of just one, as shown in the image below.

Image

### Motivation

This feature benefits both mouse/keyboard and controller input, with the main motivation being improved controller support. Implementing it would allow removal of [this workaround](https://github.com/Epic-Fight/epicfight/blob/be0d5a484ae96e8b195f048d9c233297016772e8/src/main/java/yesman/epicfight/compat/controlify/ControlifyCompat.java#L536-L541).

Controlify utilizes the vanilla arrow key navigation system for controller input. The existing workaround instead enforces virtual mouse behavior, causing controller navigation to mimic mouse input.

### Controlify?

> [!NOTE]
> To be clear, we don't need any Controlify-specific APIs or even a controller to test or implement this. It already handles that internally; we just need to avoid overriding `SkillEditScreenProcessor#virtualMouseBehaviour` after fixing this issue.

The Skill Tree, on the other hand, already has good enough controller navigation support.

### Current behavior

Here is a full video showcasing the issue. At first, I tried to navigate using a controller, then I closed the skill editor and opened it again while using the arrow keys this time.

https://github.com/user-attachments/assets/76293362-8d28-4b4a-9c68-3ec227ff868d

The fix shouldn't be manual by overriding `keyPressed`. All vanilla screens support this feature, so we need to adapt the vanilla behavior by trying to discover what caused the issue.

### Desired Behavior

Here is an example of what the new behavior might look like:

https://github.com/user-attachments/assets/d86de182-4184-486e-af40-e73b266af62f

### Fix

Fairly simple, replace `isHover()` with `isHoveredOrFocused()`:

```diff
- int y = ((this.isHover() || selectedSlotButton == this) && !this.skillContainer.onReplaceCooldown()) ? 35 : 17;
+ int y = ((this.isHoveredOrFocused() || selectedSlotButton == this) && !this.skillContainer.onReplaceCooldown()) ? 35 : 17;
```

Take a look at [this commit](https://github.com/GaylordFockerCN/DialogueLib/commit/43a54b0858877386d8610bb0817d48efdb13a275#diff-48e7f67abff128a29bdf4c2c6d075926e962e9c16df0a228a72d0ba66705651dL7-L30) for example.

https://github.com/user-attachments/assets/5b9f2de7-8198-4357-ab9e-9d6e787db9ef

The scrolling is still not fixed.

While I'm still unfamiliar with these APIs, we probably need to override `changeFocus` or `SlotButton#nextFocusPath` to handle the scrolling for `SkillEditScreen`:

```java
public class SkillEditScreen extends Screen {
    @Override
    public @Nullable ComponentPath nextFocusPath(@NotNull FocusNavigationEvent event) {
        // Overriding this is probably unnecessary or irrelevant
        System.out.println("nextFocusPath: " + event);
        return super.nextFocusPath(event);
    }

    @Override
    protected void changeFocus(@NotNull ComponentPath path) {
        System.out.println("changeFocus: " + path);
        super.changeFocus(path);
    }

    @Override
    protected void setInitialFocus(@NotNull GuiEventListener listener) {
        System.out.println("setInitialFocus: " + listener);
        super.setInitialFocus(listener);
    }

    class SlotButton extends Button {
        @Override
        public @Nullable ComponentPath nextFocusPath(@NotNull FocusNavigationEvent event) {
            System.out.println("[SlotButton] nextFocusPath: " + event);
            return super.nextFocusPath(event);
        }
    }
}
```

EchoEllet:
Fixes #2203

### Credit

- [yesssssman](https://github.com/yesssssman/) — implemented the equip skill buttons scrolling.  
- [P1nero](https://github.com/GaylordFockerCN/) — the fix in [their mod](https://github.com/GaylordFockerCN/DialogueLib/commit/43a54b0858877386d8610bb0817d48efdb13a275#diff-48e7f67abff128a29bdf4c2c6d075926e962e9c16df0a228a72d0ba66705651dL7-L30) helped me understand that the issue is related to focus and state checking.

### 1.20.1

#2210

### DIFF
--- a/src/main/java/yesman/epicfight/client/gui/screen/SkillEditScreen.java
+++ b/src/main/java/yesman/epicfight/client/gui/screen/SkillEditScreen.java
@@ -46,7 +46,7 @@ public class SkillEditScreen extends Screen {
 	
 	private static final int MAX_SKILL_OPTIONS_ROWS = 6;
 	private static final int MAX_SLOT_ROWS = 9;
-	private static final int STRIDE = 18;
+	private static final int STRIDE = SlotButton.SIZE;
 
 	private final Player player;
 	private final PlayerSkills skills;
@@ -295,11 +295,12 @@ public class SkillEditScreen extends Screen {
 	
 	@OnlyIn(Dist.CLIENT)
 	class SlotButton extends Button {
+        private static final int SIZE = 18;
 		private final SkillContainer skillContainer;
 		private final Component slotExplanation;
 		
 		public SlotButton(int x, int y, SkillContainer skillContainer, OnPress pressedAction, Component tooltipMessage) {
-			super(x, y, 18, 18, Component.empty(), pressedAction, Button.DEFAULT_NARRATION);
+			super(x, y, SIZE, SIZE, Component.empty(), pressedAction, Button.DEFAULT_NARRATION);
 			
 			this.skillContainer = skillContainer;
 			this.slotExplanation = tooltipMessage;
--- a/src/main/java/yesman/epicfight/client/gui/screen/SkillEditScreen.java
+++ b/src/main/java/yesman/epicfight/client/gui/screen/SkillEditScreen.java
@@ -24,6 +24,7 @@ import net.minecraft.util.Mth;
 import net.minecraft.world.entity.player.Player;
 import net.neoforged.api.distmarker.Dist;
 import net.neoforged.api.distmarker.OnlyIn;
+import org.jetbrains.annotations.NotNull;
 import yesman.epicfight.api.data.reloader.SkillReloadListener;
 import yesman.epicfight.main.EpicFightMod;
 import yesman.epicfight.network.EpicFightNetworkManager;
@@ -134,7 +135,7 @@ public class SkillEditScreen extends Screen {
 									.setActive(this.skills.getSkillContainer(skill) == null)
 								);
 								
-								widgetHeight.add(26);
+								widgetHeight.add(EquipSkillButton.SPACING);
 							});
 							
 							for (Button shownButton : this.equipSkillButtons) {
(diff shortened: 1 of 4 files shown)


src/main/java/yesman/epicfight/client/gui/screen/SkillEditScreen.java
```diff
@@ -323,18 +325,13 @@ protected void renderWidget(GuiGraphics guiGraphics, int mouseX, int mouseY, flo
...
 					this.setTooltip(Tooltip.create(Component.translatable(EpicFightMod.format("gui.%s.container_on_cooldown"), this.skillContainer.getReplaceCooldown() / 20)));
 				}
 			} else {
 				this.setTooltip(Tooltip.create(this.slotExplanation));
 			}
 		}
-		
-		@Override
-		protected boolean clicked(double mouseX, double mouseY) {
-			return super.clicked(mouseX, mouseY) && !this.skillContainer.onReplaceCooldown();
```
EchoEllet, line 336: FYI, this override is no longer needed due to this addition:

```diff
protected void renderWidget(GuiGraphics guiGraphics, int mouseX, int mouseY, float partialTick) {
+            this.active = !this.skillContainer.onReplaceCooldown();
```

PR #2223, opened 2025-11-19
[1.21.1] Refactor mod compatibility module checks

EchoEllet:
Cleans up the code for easier maintenance by removing duplicated `ModList#isLoaded` and `ICompatModule#loadCompatModule` calls, making it slightly easier to support Fabric. A small step forward.

Tested both the dedicated server and client game, with some mods installed, client-side only, and also on both sides.

As always, compare the differences by viewing each commit for an easier review (in the "Commits" tab).


Yesssssman:
This is actually a golden improvement!

EchoEllet:
I'm looking forward to more golden improvements!

PR #2233, opened 2025-11-22
[Docs] Add a link reference to Nvidia resource about skinned model

EchoEllet:
Adds a link to Nvidia mesh skinning as a reference in the GitHub repo so others can refer to it in the future.

For now, this is just to not lose track of this, but someday we should consider designing a more detailed guide, maybe with diagrams and examples too.
This could also be in the WIKI, too, though this is useful for other developers.

### DIFF
--- a/docs/entity-model.md
+++ b/docs/entity-model.md
@@ -1,0 +1,7 @@
+# Entity Model
+
+This document is a WIP that explains how the player and other entity models work in Epic Fight model.
+
+## References
+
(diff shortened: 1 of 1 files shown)


Forixaim (approved):

guivnf (approved):
Great information for new developers. I want to mention that this should also be included within the https://github.com/Epic-Fight/EpicFight-Docs repository, and i'd also advise you to file an issue tracker so we don't lose this from the update plans.

PR #2251, opened 2025-11-29
[1.21.1] Keep input APIs independent of EpicFightInputAction for better third-party support

### LINKED ISSUE #2194: [API] Improve the Epic Fight Input System public APIs stabilization
### Marked As Experimental

All public APIs under `yesman.epicfight.api.client` (introduced by #2122) are currently marked as experimental using `@ApiStatus.Experimental` annotation, with the following disclaimer in both Javadoc and the WIKI:

```md
* <b>Warning:</b> This API is currently marked as experimental.
 * This designation does not imply that the implementation is of an 'experimental' quality,
 * but rather indicates that classes, methods, and fields may be subject to renaming, relocation, or removal.
 * The Epic Fight team reserves the right to modify or completely remove any components of the API at any time,
 * without prior notice.
```

> [!NOTE]
> These APIs are currently marked as **experimental**.  
> This designation **does not imply** that the implementation is of an 'experimental' quality,  
> but rather indicates that classes, methods, and fields may be subject to renaming, relocation, or removal.  
> The Epic Fight team may introduce **breaking changes** to the existing API at any time without prior notice.

### Why?

This is because I planned to introduce some breaking changes to polish the API design by changing names, locations.

### Why do addons need this system?

While addons can add full controller support without these Epic Fight input APIs, however, that resulted in ugly hacks (e.g., https://github.com/GaylordFockerCN/EpicFight-Invincible/pull/8), as the Controlify keybind emulation approach didn't work well in all cases, so they need a more efficient system like the one we have here, which already solved common problems and is well-maintained (DRY).

This means we will have to solve these issues sooner and release an update so addons such as Sword Soaring, Invincible Lib, Battle Arts, Weapons of Miracles, and NightFall can start adapting this system as soon as possible, to deliver a good experience for controller users.

### Checklist

Here are some of the tasks that need to be done to get closer to the first stable version:

- [x] ~~Rename `EpicFightInputActions` to `EpicFightInputAction` to follow standard naming conventions (part of #2213) — I used `EpicFightInputActions` with other enums such as `SkillSlots`. However, I think we may need to change the current convention, so we should fix that in the new enum classes.~~ Fixed in #2231
- [x] Allow third-party mods to register and use custom input actions while supporting controller bindings
  - [x] Also allow checking whether different bindings share the same physical controller input without forcing addons to depend on Controlify APIs
- [x] Try to refactor to remove [this NeoForge/Forge event check](https://github.com/Epic-Fight/epicfight/blob/981d4fc14eb172b341e3c34306b24b705501e8e5/src/main/java/yesman/epicfight/api/client/input/handlers/InputManager.java#L197-L218) fully.
  - [x] Is the parameter [`interactionKeyEventCheck`](https://github.com/Epic-Fight/epicfight/blob/981d4fc14eb172b341e3c34306b24b705501e8e5/src/main/java/yesman/epicfight/api/client/input/handlers/InputManager.java#L101) really needed? 
- [x] Is `yesman.epicfight.api.client.input.handlers` an appropriate package name? What about the structure and files?
- [x] Move internal APIs outside of the `api` package, such as `DiscreteInputActionTrigger` and `InputUtils`.
- [x] Add a new `InputManager#getInputMode` method that is always not null and defaults to `InputMode#KEYBOARD_MOUSE` when a controller mod is not installed.
- [x] Remove unused or confusing methods from [ControllerBinding](https://github.com/Epic-Fight/epicfight/blob/b48b808d44e1fdd942a0f39c15a38eaed764dbfb/src/main/java/yesman/epicfight/api/client/input/controller/ControllerBinding.java#L15)

- [x] Confirm whether [`ControlBinding.getInputType`](https://github.com/Epic-Fight/epicfight/blob/b48b808d44e1fdd942a0f39c15a38eaed764dbfb/src/main/java/yesman/epicfight/api/client/input/controller/ControllerBinding.java#L23-L63) is a misleading API or not.

     I have added it for addons BUT it's not used by Epic Fight, so removing it won't change the behavior, and won't require any changes to any other files except [`ControlifyCompat`](https://github.com/Epic-Fight/epicfight/blob/d59b72f46f84ebd5f2903f0ce0badf8bc9280aff/src/main/java/yesman/epicfight/compat/controlify/ControlifyCompat.java#L52). Even if the whole [`IEpicFightControllerMod`](https://github.com/Epic-Fight/epicfight/blob/b48b808d44e1fdd942a0f39c15a38eaed764dbfb/src/main/java/yesman/epicfight/api/client/input/controller/IEpicFightControllerMod.java#L33) design is flawed, we need to just update [`InputManager`](https://github.com/Epic-Fight/epicfight/blob/ae4afb5eabe80e18f514fbabd04aaf78f2cab167/src/main/java/yesman/epicfight/api/client/input/handlers/InputManager.java#L34) and [`ControlifyCompat`](https://github.com/Epic-Fight/epicfight/blob/d59b72f46f84ebd5f2903f0ce0badf8bc9280aff/src/main/java/yesman/epicfight/compat/controlify/ControlifyCompat.java#L52), and that's all.

- [x] Confirm whether the function name `IEpicFightControllerMod.isBoundToSameButton` is misleading, since it doesn't just apply to controller buttons.

- [x] **Important:** Ensure the behavior consistency between mouse/keyboard and controllers when using `InputManager#isActionActive`, `InputManager#triggerOnPress`, and the others...
  - [x] For example, test it when a screen, GUI, or tooltip is visible, or in-game. Since issue #2196 occurs only when using the mouse/keyboard, this suggests the internal handling of `InputManager` may be inconsistent. **This affects the user-facing behavior, not just the API design.**
- [x] Migrate all doc comments from Java-HTML to Markdown to adapt Java 23+
- [x] Move all Minecraft input Actions out of `EpicFightInputAction` and remove `isVanilla` and update `ControlifyCompat` to adapt
- [x] Extract classes out of `ControlifyCompat` and rename `ControlifyCompat`

EchoEllet:
Closes #2194 

- Allows third-party controller mods to compare or check whether different controller bindings refer to the same physical control without force, depending on Controlify APIs.
- Treats all input actions equally, without depending on `EpicFightInputAction` for better addon support.
- Separates all Minecraft vanilla standard input actions from Epic Fight-specific ones, with two different enums

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -13,11 +13,11 @@
 - An option that you can always activate the TPS perspective, which was only activated when aiming with ranged weapons.
 - An option screen where you can set up the camera position in TPS perspective
 - An auto-tracking functionality that aligns the player's look to the crosshair when aiming and striking entities.
-- A new keybind that moves camera freely while locking on any entity to search another target
+- A new keybind that moves the camera freely while locking on any entity to search another target
 - A lock-on snapping feature that cycles lock-on entities in the screen by snapping mouse left or right
 - An auto-targeting functionality that searches a next target when the current lock-on entity is dead
 - An option to toggle lock-on snapping and auto target
-- An option to set the maximum distance that the player can focus entities
+- An option to set the maximum distance that the player can focus on entities
 - See the devlog [here](https://www.patreon.com/posts/tps-camera-and-141028682)
 - Epic Fight's TPS perspective will be automatically disabled when a conflicting mod, such
   as [Shoulder Surfing Reloaded](https://modrinth.com/mod/shoulder-surfing-reloaded)
@@ -37,14 +37,13 @@
 - Removed AirSlash and its related fields (SkillCategory, SkillSlot) to merge air slash and combo attacks as one skill
 - Updated the experimental Epic Fight's input API to support using custom input actions that are not a
   `EpicFightInputAction`.
+- Extracted vanilla input actions from `EpicFightInputAction` into `MinecraftInputAction`.
   [#2194](https://github.com/Epic-Fight/epicfight/issues/2194)
-
-### For Devs
-
+  [#2194](https://github.com/Epic-Fight/epicfight/issues/2194)
 - New API feature: Hooks
-    - Replaces the mod-loader event system into Epic Fight API, as we planning to support multi-loader developer environment
-    - The feautre is still WIP, supporting only events for EpicFightCameraAPI
-    - We will eventually replace all Forge/Neoforge events owned by Epic Fight into Hooks
+    - Replace the mod-loader event system into Epic Fight API, as we're planning to support multi-loader developer environment
+    - The feature is still WIP, supporting only events for EpicFightCameraAPI
+    - We will eventually replace all Forge/NeoForge events owned by Epic Fight into Hooks
 
 ## [21.13.5] - 2025-11-12
 
--- a/src/main/java/yesman/epicfight/api/client/input/InputManager.java
+++ b/src/main/java/yesman/epicfight/api/client/input/InputManager.java
@@ -11,13 +11,13 @@ import net.minecraft.client.KeyMapping;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.player.Input;
 import net.minecraft.client.player.LocalPlayer;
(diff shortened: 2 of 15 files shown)


EchoEllet:
1.20.1 PR: #2252

PR #2254, opened 2025-11-30
[1.21.1] Fixed the animation looping error, caused by unprotected elapsed time variable between 0 ~ total time

Yesssssman:
Initially reported by @squoshi, a regression driven by the binary search and baked keyframes feature in 20.12.8

https://github.com/Epic-Fight/epicfight/commit/9e2519390e3fb358155c058ca3290e1fa532c1e8

PR #2255, opened 2025-11-30
[1.20.1] Fixed the animation looping error, caused by unprotected elapsed time

Yesssssman:
same goes with #2254, but 1.20.1.


guivnf (approved):
Perfectly resonates the changes applied to 1.21

PR #2256, opened 2025-11-30
[1.20.1] Fixed the dedicated server crash when initializing the server

Yesssssman:
Using `LocalPlayer` in mixin code causes a dedicated server crash, even tho it's guarded by type-checking.

```JAVA
private float epicfight$serverAiStep(Player player) {
		if (player.isLocalPlayer()) {
			return EpicFightCameraAPI.getInstance().getYRotForHead((LocalPlayer)player);
		}
		
		return player.getYRot();
    }
```

So moved the type-casting into EpicFightCameraAPI

```JAVA
@ApiStatus.Internal
	public float getYRotForHead(Player player) {
		if (!player.isLocalPlayer()) {
			// Casting player to LocalPlayer led to a crash in dedicated server, so we delegated type checking to here
			// See MixinLivingEntity
			throw new IllegalArgumentException("Only LocalPlayer are allowed to this parameter");
		}
		
		return (this.isTPSMode() && (Mth.abs(Mth.wrapDegrees(this.cameraYRot - player.yBodyRot)) <= 51.0F || this.predicateCouplingPlayer())) ? this.cameraYRot : player.getYRot();
	}
```

### DIFF
--- a/src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
+++ b/src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
@@ -901,8 +901,6 @@ public final class EpicFightCameraAPI {
 	@ApiStatus.Internal
 	public float getYRotForHead(Player player) {
 		if (!player.isLocalPlayer()) {
-			// Casting player to LocalPlayer led to a crash in dedicated server, so we delegated type checking to here
-			// See MixinLivingEntity
 			throw new IllegalArgumentException("Only LocalPlayer are allowed to this parameter");
 		}
 		
--- a/src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
+++ b/src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
@@ -896,11 +896,17 @@ public final class EpicFightCameraAPI {
 	}
 	
 	/**
-	 * REturns a new basis for {@link LivingEntity#yRotHead} instead of coupling it to {@link Entity#getYRot}
+	 * Returns a new basis for {@link LivingEntity#yRotHead} instead of coupling it to {@link Entity#getYRot}
 	 */
 	@ApiStatus.Internal
-	public float getYRotForHead(LocalPlayer localPlayer) {
-		return (this.isTPSMode() && (Mth.abs(Mth.wrapDegrees(this.cameraYRot - localPlayer.yBodyRot)) <= 51.0F || this.predicateCouplingPlayer())) ? this.cameraYRot : localPlayer.getYRot();
+	public float getYRotForHead(Player player) {
+		if (!player.isLocalPlayer()) {
+			// Casting player to LocalPlayer led to a crash in dedicated server, so we delegated type checking to here
+			// See MixinLivingEntity
+			throw new IllegalArgumentException("Only LocalPlayer are allowed to this parameter");
+		}
+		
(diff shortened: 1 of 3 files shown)


src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
```diff
@@ -896,11 +896,17 @@ public void fireCameraBuildPost(Camera camera, float partialTick) {
...
-	 * REturns a new basis for {@link LivingEntity#yRotHead} instead of coupling it to {@link Entity#getYRot}
+	 * Returns a new basis for {@link LivingEntity#yRotHead} instead of coupling it to {@link Entity#getYRot}
 	 */
 	@ApiStatus.Internal
-	public float getYRotForHead(LocalPlayer localPlayer) {
-		return (this.isTPSMode() && (Mth.abs(Mth.wrapDegrees(this.cameraYRot - localPlayer.yBodyRot)) <= 51.0F || this.predicateCouplingPlayer())) ? this.cameraYRot : localPlayer.getYRot();
+	public float getYRotForHead(Player player) {
+		if (!player.isLocalPlayer()) {
+			// Casting player to LocalPlayer led to a crash in dedicated server, so we delegated type checking to here
+			// See MixinLivingEntity
```
EchoEllet: I have a suggestion:

```md
/// This method takes a [Player] instead of [LocalPlayer] because casting
/// to the client-only [LocalPlayer] inside a mixin (e.g., in [LivingEntity])
/// would crash a dedicated server due to Forge's `@OnlyIn(Dist.CLIENT)`.
```

EchoEllet (changes requested):

src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
```diff
@@ -896,11 +896,17 @@ public void fireCameraBuildPost(Camera camera, float partialTick) {
 	}
 	
 	/**
-	 * REturns a new basis for {@link LivingEntity#yRotHead} instead of coupling it to {@link Entity#getYRot}
+	 * Returns a new basis for {@link LivingEntity#yRotHead} instead of coupling it to {@link Entity#getYRot}
 	 */
```
EchoEllet: Just one more change:

```suggestion
/// Returns a new basis for [LivingEntity#yRotHead] instead of coupling it to [Entity#getYRot].
///
/// This method takes a [Player] instead of [LocalPlayer] because casting
/// to the client-only [LocalPlayer] inside a mixin (e.g., in [LivingEntity])
/// would crash a dedicated server due to Forge's `@OnlyIn(Dist.CLIENT)`.
```

src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
```diff
@@ -896,11 +896,17 @@ public void fireCameraBuildPost(Camera camera, float partialTick) {
...
 	@ApiStatus.Internal
-	public float getYRotForHead(LocalPlayer localPlayer) {
-		return (this.isTPSMode() && (Mth.abs(Mth.wrapDegrees(this.cameraYRot - localPlayer.yBodyRot)) <= 51.0F || this.predicateCouplingPlayer())) ? this.cameraYRot : localPlayer.getYRot();
+	public float getYRotForHead(Player player) {
+		if (!player.isLocalPlayer()) {
+			// Casting player to LocalPlayer led to a crash in dedicated server, so we delegated type checking to here
```
EchoEllet:
```suggestion

```

src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
EchoEllet:
```suggestion

```

EchoEllet (approved):

PR #2260, opened 2025-11-30
[1.21.1] Generate -api.jar with public APIs only

EchoEllet:
Configures Gradle to generate API JAR file (with the sources JAR file) while following Java standards.

There is a way to ship the sources in one JAR file instead of 2, but this is non-standard and may not display the source in some IDEs.

image

image

image

While our public API is quite stable yet, we should make this step early to allow consumers to adapt to the API earlier.

When the API becomes more stable, we won't need any changes in Gradle scripts.


EchoEllet:
This can be backported to `1.20.1`, but since we will be discounting the full support soon, only some features and bug fixes will be backported, so the API will already be stable enough to develop addons with no more breaking changes.

Yesssssman (approved):

PR #2262, opened 2025-11-30
[1.21.1] Adapt 'Keep a Changelog'

EchoEllet:
- Renamed `RELEASE_NOTES.md` -> `CHANGELOG.md`
  - Usually, `RELEASE_NOTES.md` contains the release notes for a version, while `CHANGELOG.md` for all versions, and `CHANGELOG.md` is commonly adopted in many projects, even outside of JVM and MC modding.
- Updated Gradle scripts to produce an error instead of a warning when the version is not found in `CHANGELOG.md` and changed the current version to `21.13.5` (since that is the current latest published version):
  - The version should be changed only once, when we publish the unreleased changes. Which means, update `mod_version` to `21.14.0` (in `gradle.properties` and `[Unreleased]` to `[21.14.0] - XXXX-YY-ZZ` (in `CHANGELOG.md`)
  - We should not change the `mod_version` as a way to update JAR file names in `build/libs` to upload them somewhere else. There are better alternatives (e.g., `./gradlew build -Pmod_version=1.2.3`).
- Updated the changelog file content to follow the common consistent conventions.

This is what [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) says about the changelog file name:

> What should the changelog file be named?
Call it CHANGELOG.md. Some projects use HISTORY, NEWS or RELEASES.
> 
> While it's easy to think that the name of your changelog file doesn't matter that much, why make it harder for your end users to consistently find notable changes?

This prepares for the next update, which is to use git version tags, and some GitHub actions for a better automated yet efficient workflow.


Yesssssman:
> We should not change the mod_version as a way to update JAR file names in build/libs to upload them somewhere else. There are better alternatives.

Is there a specific alternative for this? I actually use it to share builds in private, to the addon developers like @Reascer, P1nero but without pushing the version change.

If it's fine to change the version in `gradle.properties` in local side, then I'm fine.

EchoEllet:
> If it's fine to change the version in `gradle.properties` in local side, then I'm fine.

It's fine as long as you don't push changes to `CHANGELOG.md` and `gradle.properties`, but there is a better alternative:

```
./gradlew build -Pmod_version=1.2.3
```

image

This changes the version just for this run without touching `mod_version` in `gradle.properties`.

PR #2265, opened 2025-12-01
[1.20.1] Fixed Lock-on stuck

Yesssssman:
While testing PvP with camera features on a dedicated server, I and @EchoEllet found a lock-on malfunction when a focusing entity is too close.

**Note the player is trying to move left**

https://github.com/user-attachments/assets/67a5c364-b04d-410c-8619-1a7261b4eba5

This is because I tried to prevent the infinite turning issue when the crosshair is never reachable to the focusing target.

**Infinite Looping issue**

https://github.com/user-attachments/assets/e4b83737-231e-4260-9339-aa4bcc34756a

So we had to find an alternative way, we searched how Better Lock On + Shoulder Surfing Reloaded work together and found the basis location of lock-on is the player's eye position, not a camera position.

**BUT** I thought this solution is a bit out of the innate of Lock-on functionality, so I decided to mix these cases.

In **EpicFightCameraAPI**, line 711
```java
double toTargetDistanceSqr = localPlayer.position().distanceToSqr(this.focusingEntity.position());
					
// Lerp the start and end location of the camera arm for lock-on based on the distance between the player and the focusing entity
lockStart = MathUtils.lerpVector(localPlayer.getEyePosition(), cameraPos, (float)Mth.clampedMap(toTargetDistanceSqr, 1.0F, 18.0F, 0.2F, 1.0F));
lockEnd = MathUtils.lerpVector(this.focusingEntity.getEyePosition(), this.focusingEntity.getBoundingBox().getCenter(), (float)Mth.clampedMap(toTargetDistanceSqr, 0.0F, 18.0F, 0.5F, 1.0F));
```

This code **interpolates** the beginning position of lock-on from the player's eye to the camera. In other words, the lock-on will be based on the camera location when it's far enough to locate the focusing entity in the center of the screen, while it will be based on player's eye location when the distance is too close to prevent the **infinite turning** I mentioned above.

https://github.com/user-attachments/assets/53bc89b6-7d78-40ef-99e7-ad414083048e

In this video, you can see the player is moving on the left side circling around the focusing entity unlike the first video.

### DIFF
--- a/src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
+++ b/src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
@@ -326,11 +326,8 @@ public final class EpicFightCameraAPI {
 		Vec3 cameraLocation = this.minecraft.gameRenderer.getMainCamera().getPosition();
 		
 		// Create a compact projection matrix without view, hurt bob
-		PoseStack posestack = new PoseStack();
-		double fov = this.minecraft.gameRenderer.getFov(this.minecraft.gameRenderer.getMainCamera(), 1.0F, true);
+		Matrix4f compactProjection = this.getCompactProjectionMatrix();
 		double lockOnRange = this.getFocusingEntityPickRange();
-		posestack.mulPoseMatrix(this.minecraft.gameRenderer.getProjectionMatrix(fov));
-		Matrix4f compactProjection = posestack.last().pose();
 		
 		// Select the nearest target on the screen from the given direction
 		Optional<Pair<LivingEntity, Float>> next = entitiesInLevel.stream()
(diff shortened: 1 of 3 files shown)


src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
```diff
@@ -355,6 +352,17 @@ public boolean setNextLockOnTarget(int direction) {
 		return next.isPresent();
 	}
 	
+	/**
+	 * Creates a compact projection matrix without view, hurt bob
+	 * @return
+	 */
+	private Matrix4f getCompactProjectionMatrix() {
```
EchoEllet, line 359: What is "hurt bob"?

> * @return

This does not seem to have any docs, should we remove it?

EchoEllet (approved):
LGTM. Thank you for the detailed explanation.

src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
Yesssssman: the camera shake effect when the player get hurt

Yesssssman: > > * @return
> 
> This does not seem to have any docs, should we remove it?

yes

PR #2266, opened 2025-12-01
[1.21.1] Fixed Lock-on stuck

Yesssssman:
Same as #2265 but in 1.21.1

### DIFF
--- a/src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
+++ b/src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
@@ -323,11 +323,8 @@ public final class EpicFightCameraAPI {
         Vec3 cameraLocation = this.minecraft.gameRenderer.getMainCamera().getPosition();
 
         // Create a compact projection matrix without view, hurt bob
-        PoseStack posestack = new PoseStack();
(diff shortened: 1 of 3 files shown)


EchoEllet (approved):
Just a minor detail.

src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
```diff
@@ -355,6 +352,16 @@ public boolean setNextLockOnTarget(int direction) {
         return next.isPresent();
     }
 
+    /**
+     * Creates a compact projection matrix without view, hurt bob
+     */
```
EchoEllet: Nit: Consider using markdown comments as we're migrating from HTML comments:

```suggestion
     /// Creates a compact projection matrix without view, hurt bob
```

src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
```diff
@@ -355,6 +352,16 @@ public boolean setNextLockOnTarget(int direction) {
         return next.isPresent();
     }
 
+    /**
+     * Creates a compact projection matrix without view, hurt bob
```
EchoEllet: > Creates a compact projection matrix without view, hurt bob

QQ: What does "hurt bob" mean?

PR #2272, opened 2025-12-01
[1.21.1] Workaround AzureLib breaking changes and add reusable version checking

### LINKED ISSUE #2269: Disable AzureLib compatibility module for versions older than 3.X
Fix [this comment](https://github.com/Epic-Fight/epicfight/pull/2142#discussion_r2474222706), and disable the compatibility module.

EchoEllet:
Adds reusable mod dependency version parsing and checking in a standard way, and uses that to disable AzureLib compatibility modules for versions 3.X.X and newer versions.

More Gradle script improvements and use official Maven repositories for GeckoLib, AzureLib, and AzureLibArmor instead of Curse Forge, to support transitive dependencies, sources JAR file, readable version String, better reliability, and support for modules (`neo`, `common`, `fabric`), which is critical for multi-mod loader support.

Fixes #2269

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -29,3 +29,6 @@ kubejs_version=2101.7.1-build.181
 rhino_version=2006.2.4-build.17
 architectury_version=13.0.8
 controlify_version=2.4.3+1.21.1-neoforge
+azurelib_version=2.3.28
+azurelib_armor_version=2.3.15
+geckolib_bersion=4.8.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **Fixed:** The **Shoulder Surfing compatibility module** was not being registered,
   causing issues with its intended functionality.
 - Fixed the mining crosshair not to show in vanilla mode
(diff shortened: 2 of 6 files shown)


EchoEllet:
This also addresses [this comment](https://github.com/Epic-Fight/epicfight/pull/2142#discussion_r2474222706).

This can't be cherry-picked to `1.20.1` as we need to backport other `1.2.1.1` changes to `1.20.1` first, which we haven't.
We can, however, continue using the [workaround in this PR](https://github.com/Epic-Fight/epicfight/pull/2142#discussion_r2474222706), which is less modular, not reusable, and does not support AzureLibArmor.

But that's fine as long as we don't override `1.21.1` changes by porting `1.20.1` changes over `1.21.1`.

gradle.properties
```diff
@@ -29,3 +29,6 @@ kubejs_version=2101.7.1-build.181
 rhino_version=2006.2.4-build.17
 architectury_version=13.0.8
 controlify_version=2.4.3+1.21.1-neoforge
+azurelib_version=2.3.28
+azurelib_armor_version=2.3.15
+geckolib_bersion=4.8.2
```
EchoEllet, line 34: There is a typo in here, `version`. I will fix it tomorrow as it's sleep time for me.

Yesssssman (approved):

PR #2276, opened 2025-12-02
[1.20.1] Added TPS activate event, modified BuildCameraTransform

Yesssssman:
Added a new event to control the vanilla camera process

In EpicFightCameraAPI, line 793
```java
public boolean setupCamera(Camera camera, float partialTick) {
		if (!camera.getEntity().is(this.minecraft.player)) {
			return false;
		}
		
		BuildCameraTransform.Pre event = new BuildCameraTransform.Pre(this, camera, partialTick);
		EpicFightClientHooks.Camera.BUILD_TRANSFORM_PRE.post(event);
		
		if (event.hasCanceled()) {
			return event.cancelVanillaCameraSetup(); // Exposed the event to control the return
		}
...
```

This allows the event to completely turn off the Epic Fight modification and let Minecraft handle the camera setup. The method return value is used by `MixinCamera`

```java
public void epicfight$setup(BlockGetter level, Entity entity, boolean detached, boolean thirdPersonReverse, float partialTick, CallbackInfo callbackInfo) {
	this.initialized = true;
	this.level = level;
	this.entity = entity;
	this.detached = detached;
	
	EpicFightCameraAPI cameraApi = EpicFightCameraAPI.getInstance();
	Camera camera = (Camera)(Object)this;
	
	if (cameraApi.setupCamera(camera, partialTick)) { // where `setupCamera` is called
		callbackInfo.cancel();
	}
	
	cameraApi.fireCameraBuildPost(camera, partialTick);
}
```

Additionally, a new event to control TPS camera activation has been introduced. `IsTPSMode` is widely used to check Epic Fight camera modification, so this event allows turning off all TPS camera features at once.

To demonstrate the action, I created an event hook that should disable TPS camera when the player is using an item. Without the new activation event, this was our trial to disable Epic Fight TPS (with Shoulder Surfing Reloaded, by @EchoEllet)

```java
EpicFightClientHooks.Camera.BUILD_TRANSFORM_PRE.registerEvent(event -> {
        if (Minecraft.getInstance().player != null && Minecraft.getInstance().player.isUsingItem()) {
        	event.setCancelVanillaCameraSetup(false);
        	event.cancel();
        }
});
```

This might lead to misfunctionality since `IsTPSMode()` still returns true and the mouse input is still being interrupted.

https://github.com/user-attachments/assets/594d05c1-643c-459d-9865-d2473055e9e3

The new event completely solved this issue.

```java
EpicFightClientHooks.Camera.ACTIVATE_TPS_CAMERA.registerEvent(event -> {
    if (Minecraft.getInstance().player != null && Minecraft.getInstance().player.isUsingItem()) {
        event.cancel();
    }
});
```

https://github.com/user-attachments/assets/5aae4fbd-2b27-4055-b88a-1037b0f9c0f1


Yesssssman:
This is unrelated to the camera change. I changed Java class paths in .dll library to reflect package name update in [this commit](https://github.com/Epic-Fight/epicfight/pull/2264)

src/main/java/yesman/epicfight/api/client/event/types/BuildCameraTransform.java
```diff
@@ -23,9 +23,26 @@ public float getPartialTick() {
...
 			super(cameraApi, camera, partialTick);
 		}
+		
+		/**
+		 * Set true if you want to disable the vanilla camera setup process
+		 * This means, you have modified the camera transform through this event
+		 * and don't want it to be overwrite by vanilla camera setups.
+		 * 
+		 * @see MixinCamera#epicfight$setup
+		 */
```
EchoEllet:
```suggestion
		/// Set to `true` to disable the vanilla camera setup process.
		/// Indicates that the camera transform has been modified through this event
		/// and should not be overwritten by vanilla camera setups.
		///
		/// @see MixinCamera#epicfight$setup
```

We should keep any new comments in Markdown to ease the migration in the future.

I have also made some other changes to keep the doc comments slightly easier to read and formal, without saying `You need to` or `we have to` or `but this could`.

src/main/java/yesman/epicfight/api/event/EventHook.java
```diff
@@ -45,31 +45,31 @@ public boolean post(T eventInstance) {
...
 	}
 	
 	/**
 	 * Register an event with default name
 	 * @param priority determines the order of the event in descending order
 	 */
-	public void registerPassiveEvent(DefaultEventSubscription<T> subscription, int priority) {
-		this.registerPassiveEvent(subscription, getDefaultSubscriberName(), priority);
+	public void registerEvent(DefaultEventSubscription<T> subscription, int priority) {
+		this.registerEvent(subscription, getDefaultSubscriberName(), priority);
```
EchoEllet, line 57: Good change!
To be honest, I was hesitant about `registerPassiveEvent`

src/main/java/yesman/epicfight/api/client/event/types/BuildCameraTransform.java
Yesssssman: For now, I just kept old comment style in 1.20.1 since I use eclipse and others all use javadoc style comments.

I uses markdown standard only in 1.21.1.

EchoEllet: > For now, I just kept old comment style in 1.20.1 since I use eclipse and others all use javadoc style comments.

I think that's fine, the doc comment is still readable, and we should not wait for the current JDK to be exactly `23` to make it easier to handle the change.

Does the doc comment not get displayed in Eclipse?

EchoEllet (approved):
LGTM, still need more testing with Shoulder Surfing though

src/main/java/yesman/epicfight/api/client/event/types/BuildCameraTransform.java
```diff
@@ -23,9 +23,24 @@ public float getPartialTick() {
...
+		///
+		/// @see MixinCamera#epicfight$setup
+		public void setCancelVanillaCameraSetup(boolean flag) {
+			this.cancelVanillaCameraSetup = flag;
+		}
+		
+		public boolean cancelVanillaCameraSetup() {
+			return this.cancelVanillaCameraSetup;
+		}
 	}
```
EchoEllet, line 45: I think `cancelVanillaCameraSetup` can be confused as if it has a side effect or cancels something, but it just returns `cancelVanillaCameraSetup`.

Image

My suggestion is to refactor to `isCancelVanillaCameraSetup` or any other name.

Yesssssman:
NOTE: This branch will be used to backport SSR compatibility events from 1.21.1

Issue #2258, opened 2025-11-30
Add lock-on support for Shoulder Surfing


EchoEllet:
### **Long-story short**

- [x] Update Shoulder Surfing to not mixin on Epic Fight (https://github.com/Exopandora/ShoulderSurfing/pull/374)
- [x] Add explicit lock-on support in [`ShoulderSurfingCompat`](https://github.com/Epic-Fight/epicfight/blob/bb785325d423e1b8f305779ddd49b36525391334/src/main/java/yesman/epicfight/compat/shouldersurfing/ShoulderSurfingCompat.java).

### Details

Shoulder Surfing is a mod that adds enhanced TPS support with features such as camera decoupling.

The mod added lock-on support in [this commit](https://github.com/Exopandora/ShoulderSurfing/commit/f1f4f472ff9ac73ef6f724e451af433b59cf1d0b) (the [latest mixin](https://github.com/Exopandora/ShoulderSurfing/blob/master/common/src/main/java/com/github/exopandora/shouldersurfing/compat/mixins/epicfight/MixinRenderEngine.java)).

Epic Fight has added camera decoupling and enhanced TPS support, too, and using both would result in undefined behavior, so I sent a PR #2224 to disable Epic Fight's TPS when "Shoulder Surfing" or "Better third-person" is installed.

This still does not fix the issue, because Shoulder Surfing mixins on `RenderEngine`, which has been updated, and I [added deprecated methods](https://github.com/Epic-Fight/epicfight/pull/2224/files#diff-857904bf46a6db49080264e24ef789f2793bf3e204fa85859e73dc65ddf1d944R375-R391) that do nothing just to avoid a runtime crash.

~~Then I sent [a PR to Shoulder Surfing](https://github.com/Exopandora/ShoulderSurfing/pull/369) to remove these mixins,~~ so we handle lock-on support from Epic Fight's side, to avoid future breakage.

### Lock-on fix

Cancel `EpicFightCameraAPI#setupCamera` via an Epic Fight event (e.g., `EpicFightClientHooks.Camera.BUILD_TRANSFORM_PRE`) if perspective is Shoulder Surfing, and then add explicit support for lock-on.

### Related issues

- Related https://github.com/Exopandora/ShoulderSurfing/issues/359
- Related https://github.com/Exopandora/ShoulderSurfing/issues/315
- Related https://github.com/Exopandora/ShoulderSurfing/issues/239

EchoEllet:
> Lock-on fix
[...]

PR #2277, opened 2025-12-02
[1.21.1] Added TPS activate event, modified BuildCameraTransform

Yesssssman:
Same goes with #2276, but in 1.21.1

PR #2283, opened 2025-12-02
Improve compatibility with Shoulder Surfing Reloaded

Exopandora:
This PR aims to improve compatibility with Shoulder Surfing Reloaded (SSR) and serves more of a demo purpose. The first two commits demonstrate what is necessary to get lock-on working again (although the crosshair alignment could be better).

What still needs to be done:
1. The code should be moved to a different location if possible, and proper runtime checks should be added, whether SSR is actually installed.
2. The lock-on raytrace should be adjusted to use the object picker provided by the SSR API.

Related to #2258

### DIFF
--- a/src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
+++ b/src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
@@ -1,5 +1,7 @@
 package yesman.epicfight.api.client.camera;
 
+import com.github.exopandora.shouldersurfing.api.client.IShoulderSurfing;
+import com.github.exopandora.shouldersurfing.api.client.ShoulderSurfing;
 import com.mojang.blaze3d.vertex.PoseStack;
 import com.mojang.datafixers.util.Pair;
 import net.minecraft.client.Camera;
@@ -20,7 +22,11 @@ import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.item.UseAnim;
 import net.minecraft.world.level.ClipContext;
 import net.minecraft.world.level.block.state.BlockState;
-import net.minecraft.world.phys.*;
+import net.minecraft.world.phys.AABB;
+import net.minecraft.world.phys.BlockHitResult;
+import net.minecraft.world.phys.EntityHitResult;
+import net.minecraft.world.phys.HitResult;
+import net.minecraft.world.phys.Vec3;
 import net.neoforged.neoforge.client.event.ViewportEvent;
 import net.neoforged.neoforge.entity.PartEntity;
 import org.apache.commons.lang3.mutable.MutableBoolean;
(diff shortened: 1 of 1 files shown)


Exopandora:
This PR also fixes general camera rotations when TPS is disabled and lock-on is active.

EchoEllet:
Thanks for crafting this PR! I wasn't expecting it.

> The code should be moved to a different location if possible, 

We have some WIP events in #2277, and we will add more to get Shoulder Surfing working fully using Epic Fight's public API, without mixins or internal implementation details this time, to prevent future breakage (hopefully!).

And then use the new events in [`ShoulderSurfingCompat`](https://github.com/Epic-Fight/epicfight/blob/bb785325d423e1b8f305779ddd49b36525391334/src/main/java/yesman/epicfight/compat/shouldersurfing/ShoulderSurfingCompat.java#L49).

I will look into this in detail soon.

Yesssssman (approved):

Yesssssman:
Thank you very much! We will adapt this change to our event API system, keeping your contribution in the commit log.

EchoEllet:
> The lock-on raytrace should be adjusted to use the object picker provided by the SSR API.

Is the object picker provided by SSR an internal or public API?

Exopandora:
> > The lock-on raytrace should be adjusted to use the object picker provided by the SSR API.
> 
> Is the object picker provided by SSR an internal or public API?

It’s public API. I can create a second PR later if needed. For reference, see https://github.com/Exopandora/ShoulderSurfing/blob/master/common/src/main/java/com/github/exopandora/shouldersurfing/compat/mixins/tslatentitystatus/MixinTESHud.java#L39-L51

EchoEllet:
@Yesssssman accidentally merged your PR and missed the "demo" part.

Should we adapt the changes by using Epic Fight events (in `ShoulderSurfingCompat`), or do you plan to make more changes, since it was a draft PR?

> The lock-on raytrace should be adjusted to use the object picker provided by the SSR API.

Though what is the reason? Is it for compatibility with other SSR mods, or to work better with other related features?

Exopandora:
> Though what is the reason? Is it for compatibility with other SSR mods, or to work better with other related features?

Depending on the configuration and the players state, the object picker will start the trace from different locations (e.g. player head or camera), or adjust the length of the ray trace. This ensures that there will be no misalignment between the visual crosshair, and what the game actually highlights.

Exopandora:
After some testing, it does not seem to be necessary to use the object picker of Shoulder Surfing.

EchoEllet:
> After some testing, it does not seem to be necessary to use the object picker of Shoulder Surfing.

I see, thanks for confirming! Because I tried different configs and couldn't reproduce an issue.

It's just the [initial camera shake](https://github.com/Epic-Fight/epicfight/pull/2289#issuecomment-3607445136) that's an obvious user-facing issue.

Exopandora:
I'll have to revoke my previous statement. You can still break it when you set `crosshair_type` to `DYNAMIC`. To make it more obvious, additionally set `object_picker.pick_vector` to `PLAYER`. Though, I do not think that it is an huge issue and most players will not even notice.

EchoEllet:
> You can still break it when you set crosshair_type to DYNAMIC. To make it more obvious, additionally set

BTW, Epic Fight has also added support for custom crosshair rendering.

Should we always disable it as well when Shoulder Surfing Reloaded is installed?

Or only when `crosshair_type` is not set to `none` with a tooltip in Epic Fight's config screen?

Exopandora:
I'm not sure what "support for custom crosshair rendering" means exactly in this case, but SSR only offsets the crosshair, so it _should_ not be an issue.

EchoEllet:
> I'm not sure what "support for custom crosshair rendering" means exactly in this case,

Epic Fight patches the vanilla Crosshair HUD to be contextual:

image

Since the recent updates, it's no longer needed to switch between battle and mining/vanilla modes manually via R key; now it depends on the context, so if the crosshair is aiming at a block, for example, but it might be a bit more complex than that:

- players can still change an Epic Fight config to revert to the old behavior; they can only break blocks in mining, but not in battle mode, even when aiming at a block.
- some other features, such as guard (I'm uncertain though).

I'm going to [ignore this for now](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it), as it seems to be minor, but if you think that it should be disabled, let me know.

PR #2284, opened 2025-12-03
[1.21.1] Fixed duplicated registration of skill data keys

Yesssssman:
Skill data keys with `syncronizeToRemotePlayers = true` will cause the duplicated data key registration crash. This is because of the new feature introduced in `20.12,x`, the skill update, where the skills are synchronized to remote clients to express some VFX effects like Adaptive Skin overlay and Berserker's particle generation

The point is, when the skill is equipped in remote clients, it registers data keys specified in their initial registration.

`SkillContainer.java:`
```java
public void setSkillRemote(@Nullable Skill skill) {

...

    Set<Holder<SkillDataKey<?>>> datakeys = SkillDataKeyCallbacks.getSkillDataKeyMap().get(skill.getClass());
			
    if (datakeys != null && !datakeys.isEmpty()) {
	    datakeys.stream().filter(holder -> holder.value().syncronizeToRemotePlayers()).forEach(this.skillDataManager::registerData);
    }
```
But the initial registration was also conducted at `SkillDataManager.java`

```java
public void onTracked(EpicFightNetworkManager.PayloadBundleBuilder bundleBuilder) {
	this.data.forEach((key, val) -> {
		if (key.value().syncronizeToRemotePlayers()) {
			SPHandleSkillData msg = new SPHandleSkillData(SPHandleSkillData.WorkType.REGISTER, this.container.getSlot(), this.container.executor.getOriginal().getId(), key);
			((SkillDataKey<Object>)key.value()).encode(msg.buffer(), val);
			
			bundleBuilder.and(msg);
		}
	});
}
```

This is the reason why other players get instantly crashed when begin to track a player with AdaptiveSkin skill, where has the only data key set `syncronizeToRemotePlayers = true`.

```java
	public static final DeferredHolder<SkillDataKey<?>, SkillDataKey<TagKey<DamageType>>> RESISTING_DAMAGE_TYPE = REGISTRY.register("resisting_damage_type", () ->
		SkillDataKey.createSkillDataKey(ByteBufCodecsExtends.tagKey(Registries.DAMAGE_TYPE), EpicFightDamageTypeTags.NONE, true, AdaptiveSkinSkill.class) // note that true third parameter is true
	)
```

The solutions are two, removing synchronization either in `onTracked` method or `setSkillRemote` method. Since `onTracked` provides the initial value of the skill too, I decided to remove the skill data key registration code from `setSkillRemote`. [See](https://github.com/Epic-Fight/epicfight/pull/2284/files#diff-90bcbcd73373c6862be2878bd4715d334dcf2783468063ea627b2023349aaabdL158).

I expect no regression because that skill data key was the only one set `syncronizeToRemotePlayers = true`.

Demonstration:
image

Now you can track the player with Adaptive Skin skill applied.

### DIFF
--- a/src/main/java/yesman/epicfight/skill/SkillContainer.java
+++ b/src/main/java/yesman/epicfight/skill/SkillContainer.java
@@ -66,9 +66,7 @@ public class SkillContainer {
 	}
 	
 	public boolean setSkill(@Nullable Skill skill, boolean initialize) {
-		/**
-		 * For remote players, call setSkillRemote instead
-		 */
+        // For remote players, call setSkillRemote instead
 		if (this.executor.isLogicalClient() && !this.executor.getOriginal().isLocalPlayer()) {
 			return false;
 		}
@@ -127,9 +125,7 @@ public class SkillContainer {
 	
 	@OnlyIn(Dist.CLIENT)
 	public void setSkillRemote(@Nullable Skill skill) {
-		/**
-		 * For server players or a local player, call setSkill instead
-		 */
+        // For server players or a local player, call setSkill instead
 		if (!this.executor.isLogicalClient() || this.executor.getOriginal().isLocalPlayer()) {
 			return;
 		}
@@ -146,7 +142,7 @@ public class SkillContainer {
 			this.skill.onRemoveClient(this);
 			this.executor.getPlayerSkills().removeSkillFromContainer(this.skill);
 		}
-		
+
 		this.skill = skill;
 		this.resetValues();
 		
@@ -154,6 +150,12 @@ public class SkillContainer {
 		this.skillDataManager.clearData();
 		
 		if (skill != null) {
+            Set<Holder<SkillDataKey<?>>> datakeys = SkillDataKeyCallbacks.getSkillDataKeyMap().get(skill.getClass());
+
+            if (datakeys != null && !datakeys.isEmpty()) {
+                datakeys.stream().filter(holder -> holder.value().syncronizeToRemotePlayers()).forEach(this.skillDataManager::registerData);
+            }
+
 			skill.onInitiateClient(this);
 			this.executor.getPlayerSkills().setSkillToContainer(skill, this);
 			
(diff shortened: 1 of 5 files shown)


src/main/java/yesman/epicfight/skill/SkillContainer.java
```diff
@@ -27,6 +22,10 @@
 import yesman.epicfight.world.capabilities.entitypatch.player.ServerPlayerPatch;
 import yesman.epicfight.world.gamerule.EpicFightGameRules;
 
+import javax.annotation.Nullable;
```
EchoEllet: @Yesssssman 

Why not use JetBrains's annotation `Nullable` since it's supported by IDEA, and widely adopted by many mods and in the Minecraft codebase.

Yesssssman: It's automatically relocated by IntelliJ. I didn't specifically choose `javax.annotation.Nullable` instead of jetbrain nullable.

There are some `javax.annotation.Nullable` imports throughout the project, we should slowly adapt this along with `@OnlyIn` annotation.

EchoEllet: > It's automatically relocated by IntelliJ. I didn't specifically choose javax.annotation.Nullable instead of jetbrain nullable.

My guess, but shouldn't we fix that for the new code additions?

Yesssssman: Sure, I just didn't notice that.

EchoEllet:
```suggestion
import org.jetbrains.annotations.Nullable;
```

Yesssssman:
Actually, there is one more consideration I didn't notice when applying this fix. We also have to consider about the situation when the player switches skills while being tracked by other players. That means, Skill data keys must be registered by this switch. Currently, `onTracked` in `SkillDataManager` is only called when entity firstly tracked by other players.

Yesssssman:
Here is the final fix(I hope). I rolled back registering keys on `setSkillRemote` method and changed work type from `SPHandleSkillData.WorkType.REGISTER` to `SPHandleSkillData.WorkType.MODIFY`. This allows syncing data keys for each case I mentioned above. Here is the demonstration.

The case when you track the player with synchronized skill data key:

https://github.com/user-attachments/assets/19afeb57-c6db-4ce3-8733-b35a9e81daf3

The case when the player switched skill:

https://github.com/user-attachments/assets/8fb0d99e-b34d-49b9-ab33-1516618e43a1

I also fixed some issues in `PlayerSkillCommand` while testing this change. Hopefully we don't have further problem or regression caused by this change.

EchoEllet (approved):

PR #2289, opened 2025-12-03
[1.21.1] Shoulder Surfing Reloaded Modular Compatibility

EchoEllet:
Based on #2283 by @Exopandora, and other commit changes by @Yesssssman.
This aims to add working lock-on support while keeping it an optional dependency.


EchoEllet:
@Exopandora this apply your changes in #2283 (most of the new changes are made by @Yesssssman), but fixes the first issue.

> The code should be moved to a different location if possible, and proper runtime checks should be added, whether SSR is actually installed.

Fixed now.

> The lock-on raytrace should be adjusted to use the object picker provided by the SSR API.

This still does not use the object picker by the Shoulder Surfing API suggested in [this comment](https://github.com/Epic-Fight/epicfight/pull/2283#issuecomment-3605913680).

Let me know if you're interested in making any suggestions to @Yesssssman's changes, or making any related changes. 

The commit history of this branch might not be well organized, so I would look at
[`ShoulderSurfingCompat`](https://github.com/Epic-Fight/epicfight/blob/7efc022d791578ffa98c716d44933b87c9382f17/src/main/java/yesman/epicfight/compat/shouldersurfing/ShoulderSurfingCompat.java#L52) for the Epic Fight X Shoulder Surfing compatibility.

Yesssssman:
https://github.com/user-attachments/assets/8e49dd6d-329f-4936-9a4a-e6f545ea7590

This is the only issue we have rn, recorded by @EchoEllet

When the player starts locking on the target by pressing the key, it has awkward camera shaking. I think we can solve this issue through the event I made [here](https://github.com/Epic-Fight/epicfight/commit/e4af81a84bb91d8429aa85cae71b6cc19c08c46f#diff-80072f8cf60c829be10c8c215766a543ab469a775c94a9725eba0aee5d36da58R93-R282).

Yesssssman (approved):

Issue #2146, opened 2025-10-29
Avoid depending on Forge/NeoForge OnlyIn annotation


EchoEllet:
Explained in [this comment](https://github.com/Epic-Fight/epicfight/pull/2141#issuecomment-3463107453), but long story short: This is an internal mod loader API that causes the game to crash in runtime whenever this class is loaded/inspected on the wrong Dist.

NeoForge Java docs says:

> This is generally meant for internal Forge and FML use only and modders should avoid its use whenever possible.

> [!TIP]
> Forge [also recommends not using it directly](https://docs.minecraftforge.net/en/latest/concepts/sides/#fmlenvironmentdist-and-onlyin), since it's only useful for some older Minecraft server JAR files where some client classes are not available on the dedicated server.
>
> See also: https://forums.minecraftforge.net/topic/86229-when-should-you-use-onlyindistdedicated_server/

It's not very useful and could cause issues with reflection, but we could instead:

1. Create two Gradle modules, one for the client and another for the server, to really force not calling any client API on the server at compile time.
2. Or, we could just document that this is a client-side only API, and it's likely already inside the `client` package, so this shouldn't be an issue.

Related #2141

EchoEllet:
This has already handled by @Yesssssman in this commit https://github.com/Epic-Fight/epicfight/commit/f082e0167835173ce25cdc0e52d742c7daa15671. However, splitting the source sets or client-only imports into a separate class might not be doable due to the breaking changes it may require.

PR #2301, opened 2025-12-07
[1.21.1] Remove OnlyIn Forge annotations for client packages

EchoEllet:
Removes most of the `OnlyIn` annotation Forge imports to clos_e #2146 (intentionally made a typo to prevent GitHub from auto-closing).

### Script

This change was automated using a script, and not manually, so the conflicts can be fixed in any branch by using the same script.

<details><summary>Script</summary>
<p>

**Disclaimer:** Generated via AI to save development time.

```bash
#!/bin/bash

find src/main/java/yesman/epicfight -name "*.java" -print0 | while IFS= read -r -d '' f; do
  if grep -q "^package .*client" "$f"; then

    # Check if @EventBusSubscriber(... value = Dist.CLIENT) exists anywhere in the file
    if grep -Eq '@EventBusSubscriber\(.*value *= *Dist\.CLIENT' "$f"; then
      KEEP_DIST_IMPORT=1
    else
      KEEP_DIST_IMPORT=0
    fi

    # Remove the OnlyIn import ALWAYS (safe)
    sed -i '' '/^import net\.neoforged\.api\.distmarker\.OnlyIn;/d' "$f"

    # Remove Dist import ONLY if not used in @EventBusSubscriber
    if [ "$KEEP_DIST_IMPORT" -eq 0 ]; then
      sed -i '' '/^import net\.neoforged\.api\.distmarker\.Dist;/d' "$f"
    fi

    # Remove @OnlyIn(Dist.CLIENT) annotation lines
    sed -i '' '/^[[:space:]]*@OnlyIn(Dist.CLIENT)/d' "$f"
  fi
done
```

</p>
</details> 

### Details

This does not remove all `@OnlyIn` annotations and related imports, but:

- only if the target class file is in a client package (e.g., `yesman.epicfight.client.world.capabilites.entitypatch.player`, `yesman.epicfight.api.client.input`), so in other words, `"client"` is part of the String.
  - There are many cases where we use `@OnlyIn(Dist.CLIENT)` in the same package as a dedicated server (common side), 
which is not something Minecraft does, and NeoForge/Fabric encourages separating client-only code from the common. Example case that does not meet this condition: `yesman.epicfight.api.utils.LevelUtil#createParticle`. **We should handle these cases manually without AI or scripts, and in a separate PR/issue.**
- remove all `Dist` imports but only if it's not used by a `@EventBusSubscriber` annotation, to prevent compile errors.
  - `OnlyIn` imports are always removed if `@OnlyIn` is removed

This makes it significantly easier to clear these imports correctly (details ##2146) and handle them via a separate client package or class, which is more recommended, has fewer dedicated crashes, and is mixin-friendly.

### Reasons

Details have already been discussed in #2146 and on Discord

**Long story short:** these are internal details and are meant only for the mod-loader to handle the cases where client classes are not shipped on the dedicated server JAR file provided by the mod-loader. All of MinecraftForge, NeoForged, and FabricMC teams recommend not using them at any cost.

image

image

image

Also, the [main maintainer of MinecraftForge strongly recommends against it](https://forums.minecraftforge.net/topic/86229-when-should-you-use-onlyindistdedicated_server/).

They block us from supporting Fabric, and also add a lot of conflicts between Forge and NeoForge branches, and make us more dependent on the mod loader, so in the future, if NeoForge gets forked, as happened before, we will need to update all these imports again, which is not future-proof at all.

There are more details to the discussion, but I have already discussed them with my team a few weeks ago.

### Testing

I have tested both the dedicated server, client and confirmed no issues; in fact, this could fix more issues (as happened with this PR https://github.com/Epic-Fight/epicfight/pull/2141#issuecomment-3463107453)

### Better alternatives?

- A separate package and class for anything that uses client-only methods or classes **(required)**
  - Sometimes, for example, in the `Animations` class, some may disagree and prefer to keep it coupled, so in these cases, we can just document the code or use `@ClientOnly` custom annotation, which just serves for documentation purposes, without bytecode runtime manipulation.
- **Bonus:** If we really want to solve the problem 100% further from core, no annotations, no hacks, no reflections, no workarounds, then it's better to separate the source sets as Fabric does (`main` and `client` instead of just `main` for both). This forces us to fix all issues at compile-time (type-safety). This is required in addition to applying the first solution.

### DIFF
--- a/src/main/java/yesman/epicfight/api/client/animation/AnimationSubFileReader.java
+++ b/src/main/java/yesman/epicfight/api/client/animation/AnimationSubFileReader.java
@@ -23,8 +23,6 @@ import com.mojang.datafixers.util.Pair;
 import net.minecraft.resources.ResourceLocation;
 import net.minecraft.server.packs.resources.Resource;
 import net.minecraft.util.GsonHelper;
-import net.neoforged.api.distmarker.Dist;
-import net.neoforged.api.distmarker.OnlyIn;
 import yesman.epicfight.api.animation.AnimationManager;
 import yesman.epicfight.api.animation.LivingMotion;
 import yesman.epicfight.api.animation.TransformSheet;
@@ -44,7 +42,6 @@ import yesman.epicfight.api.utils.ParseUtil;
 import yesman.epicfight.main.EpicFightMod;
 import yesman.epicfight.world.capabilities.entitypatch.LivingEntityPatch;
 
-@OnlyIn(Dist.CLIENT)
 public class AnimationSubFileReader {
 	public static final SubFileType<ClientProperty> SUBFILE_CLIENT_PROPERTY = new ClientPropertyType();
 	public static final SubFileType<PovSettings> SUBFILE_POV_ANIMATION = new PovAnimationType();
@@ -68,7 +65,6 @@ public class AnimationSubFileReader {
 		}
 	}
 	
-	@OnlyIn(Dist.CLIENT)
 	public static abstract class SubFileType<T> {
 		private final String directory;
 		private final AnimationSubFileDeserializer<T> deserializer;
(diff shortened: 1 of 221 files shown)


EchoEllet:
1.20.1 PR: #2302

The same script can be used to handle conflicts with any branch, so don't worry about conflicts.

guivnf (approved):
Seems like a good take, like you mentioned it has even fixed issues within dedicated servers. as well as strived to a multiloader approach, which's the path we want to advance as we grow more independant from the Forge API

Yesssssman:
@EchoEllet should we contain this change in x.14.x? Or the next API update?

Yesssssman (approved):

EchoEllet:
> should we contain this change in x.14.x? Or the next API update?

It does not affect the public API, so it's safe to publish.

PR #2304, opened 2025-12-08
[1.20.1] Camera update 1st stabilization

### LINKED ISSUE #2300: Lock-on target switch does not sync with the server
> [!NOTE]
> Originally posted by Arcadia (from the NightFall team) in [this Discord message](https://discord.com/channels/955784714264932362/1029384437760266260/1447210988049924197) (private dev channel).

The `getTarget()` method returns an outdated lock-on target and is not synced with the server.

https://github.com/user-attachments/assets/d754ab14-b0e6-4f5a-8b22-6fe17c04119a

Yesssssman:
Here are the very first issue reports from addon developers.
(Fixes #2300)

## [Lock-on target switch does not sync with the server](https://github.com/Epic-Fight/epicfight/issues/2300)

Found we're not sending a synchronization packet when switching a lock-on target via quick switch (through mouse snapping)
I added parameters for sending packets in the method that searches for a new target from the screen
```java
public boolean setNextLockOnTarget(int direction, boolean necessarilyLockingOn, boolean sendChange) {
    // ...
    if (sendChange) this.sendTargeting(this.focusingEntity);
}
```
Demonstration:

https://github.com/user-attachments/assets/73e641fe-dff7-41d1-9394-e9872ffd0933

targeting_works

I printed out the current target at the ticking method in `PlayerPatch`, only in logical server.
```java
if (!this.isLogicalClient()) System.out.println("Server Targeting: " + this.getTarget());
```

I expect this will solve the reported issue since the dev mentioned he used `getTarget` method to shot projectiles. Although, we need to fully test the usecase before we merge this pull request.
image

## Non Epic Fight patched entities don't show the target indicator
image

Fixed by not getting entitypatch through `EpicFightCapabilities.getUnparameterizedEntityPatch().ifPresent()` since the entity patch parameter for validation methods in `EntityUI` is nullable.

```java
public abstract boolean shouldDraw(LivingEntity entity, @Nullable LivingEntityPatch<?> entitypatch, LocalPlayerPatch playerpatch, float partialTicks);
```

Demonstration:
indicator_works

## Automatic Lock-on by picking the nearest entity from the crosshair

Suggested by the original author of the **Better Lock-On** addon. He recommended adding this feature so he doesn't have to update his add-on as the Epic Fight version grows.

Demonstration:

https://github.com/user-attachments/assets/8d98f3c2-9b61-44b7-a6f1-c0a98dcf0cd3

## Game crash with [Cinematic Cataclysm](https://www.curseforge.com/minecraft/mc-mods/cinematic-cataclysm)

[mclo.gs](https://mclo.gs/NcjkMms)

This is an internal Epic Fight crash for some exceptional cases.

```java
BuildCameraTransform.Pre event = new BuildCameraTransform.Pre(this, camera, partialTick);

if (!camera.getEntity().is(this.minecraft.player)) {
	event.cancel();
	return event;
}
```

Calling `event.cancel()` is prevented before they are initialized. This was originally for getting an `EventContext` instance when posting events, but this structure has a danger of leading to a `NullPointerException`. So I made `eventContext` initialized while constructing.

Also, changed `initializeEventContext` to `getEventContext` since it is not an initialization method at all. Tho I believe no breaking changes happen since it's marked as `@ApiStatus.Internal`.

### DIFF
--- a/src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
+++ b/src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
@@ -346,10 +346,6 @@ public final class EpicFightCameraAPI {
 		return ClientConfig.lockOnRange;
 	}
 	
-	/**
-	 * @Deprecated Use more parameters friendly version below
-	 */
-	@Deprecated(forRemoval = true)
 	public boolean setNextLockOnTarget(int direction) {
 		return this.setNextLockOnTarget(direction, false, true);
 	}
--- a/src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
+++ b/src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
@@ -288,23 +288,38 @@ public final class EpicFightCameraAPI {
 		if (this.lockingOnTarget == flag) {
             return;
         }
-
+		
+		boolean newlyFoundFocusingEntity = false;
+		
+		// Search a next target when trying to lock there is no focusing entity
+		if (flag && this.focusingEntity == null) {
+			newlyFoundFocusingEntity = this.setNextLockOnTarget(0, false, false);
+		}
+		
         if (!flag || this.focusingEntity != null) {
             boolean eventCanceled;
-
+            
             if (flag) {
                 LockOnEvent.Start lockOnEvent = new LockOnEvent.Start(this, this.focusingEntity);
                 EpicFightClientHooks.Camera.LOCK_ON_START.post(lockOnEvent);
                 eventCanceled = lockOnEvent.hasCanceled();
+                
+                if (eventCanceled && newlyFoundFocusingEntity) {
+                	this.focusingEntity = null;
+                }
             } else {
                 LockOnEvent.Release lockOnEvent = new LockOnEvent.Release(this, this.focusingEntity);
                 EpicFightClientHooks.Camera.LOCK_ON_RELEASED.post(lockOnEvent);
                 eventCanceled = lockOnEvent.hasCanceled();
             }
-
+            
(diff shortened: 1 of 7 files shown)


src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
```diff
@@ -331,19 +346,29 @@ public int getFocusingEntityPickRange() {
...
 	}
 	
+	/**
+	 * @Deprecated Use more parameters friendly version below
+	 */
+	@Deprecated(forRemoval = true)
+	public boolean setNextLockOnTarget(int direction) {
+		return this.setNextLockOnTarget(direction, false, true);
+	}
+
```
EchoEllet: QQ:

Should we keep this as a convenient method overload with fewer arguments required (without deprecation)?

EchoEllet (approved):
LGTM and thanks for keeping this backward compatible.
I would just keep the `setNextLockOnTarget` with fewer parameters without deprecation, to avoid redundant arguments in case they don't need it.
Of course, there might be a reason for that.

Yesssssman:
> LGTM and thanks for keeping this backward compatible. I would just keep the `setNextLockOnTarget` with fewer parameters without deprecation, to avoid redundant arguments in case they don't need it. Of course, there might be a reason for that.

Well, I don't expect developers to use the method frequently. I was also thinking of marking the method as `@ApiStatus.Internal`.

EchoEllet:
> Well, I don't expect developers to use the method frequently. I was also thinking of marking the method as `@ApiStatus.Internal`.

It seems some addon devs already used it. We should check this out in #developer on Discord.

Yesssssman:
image

Decided to leave the method as @GaylordFockerCN requested

PR #2306, opened 2025-12-08
[1.21.1] Avoid changing user's client config when SSR is installed

EchoEllet:
The previous code changes the user client config as a quick workaround to cause `EpicFightCameraAPI#isTPSMode` to return `false`.

This is inefficient and inconvenient, as IO operations are non-trivial and should be avoided for such simple compatibility.

Also, the user `IRBan` from Epic Fight Discord reported that this is causing game load crashes with some old configs when SSR is installed, so I updated the code to make use of the official supported API, for a more future-proof SSR compatibility.


EchoEllet:
We should document such changes in `CHANGELOG.md`, but the `Unreleased` section does not exist, so I avoided adding it, though the conflicts are trivial and can be solved via GitHub UI.

Yesssssman:
Good! This is exactly what we discussed at, https://discord.com/channels/955784714264932362/1302011635757486100/1446120057888309290

Yesssssman (approved):

PR #2310, opened 2025-12-08
[1.21.1] Simplify LangKeys object generation by updating MC Safe Resources

EchoEllet:
I have updated [MC Safe Resources Gradle Plugin](https://github.com/EchoEllet/mc-safe-resources-gradle) to be simpler and more convenient.
This was originally a complex Gradle task in Epic Fight that's hard to maintain, so I extracted it to a separate reusable Gradle plugin and used it in this PR #2270, and now it's even simpler, and I plan to continue simplifying it further.

This is not a breaking change, and the more manual way is still an option.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,12 +1,9 @@
-import dev.echoellet.minecraft_safe_resources.GenerateJsonKeysTask
-import dev.echoellet.minecraft_safe_resources.OutputLanguage
-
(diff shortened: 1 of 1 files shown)


EchoEllet:
This is a minor Gradle build script change and does not affect functionality or source code, so I will merge to focus on other PRs.

PR #2313, opened 2025-12-09
[1.21.1] Fixed the player being dark in inventory screen

Yesssssman:
image

Reported by @SAGESSE-CN, I confirmed the issue is only the case in 1.21.1.

Problem is coming from we're relying on `PoseStack#mulPose` which was added in 1.21.1 Minecraft. I expected this method to multiply the transform matrix by both the `pose` and `normal` matrices, but it wasn't.

```java
    Pose posestack$pose = (Pose)this.poseStack.getLast();
    posestack$pose.pose.mul(pose);
    if (!MatrixUtil.isPureTranslation(pose)) {
        if (MatrixUtil.isOrthonormal(pose)) {
            posestack$pose.normal.mul(new Matrix3f(pose));
        } else {
            posestack$pose.computeNormalMatrix();
        }
    }
```

It checks some conditions in the given matrix to apply a normal matrix. This isn't a problem for rendering entities in a world, but it is for entities in the GUI, especially the player in an inventory screen. So I switched to more reliable way, a method that directly multiplying the given matrix to `PoseStack`

```java
private static final Matrix4f BUFFER4F = new Matrix4f();
private static final Matrix3f BUFFER3F = new Matrix3f();

public static void mulStack(PoseStack poseStack, OpenMatrix4f mat) {
	OpenMatrix4f.exportToMojangMatrix(mat, BUFFER4F);
    poseStack.last().pose().mul(BUFFER4F);
    poseStack.last().normal().mul(BUFFER3F);
}
```

Demonstration:

image

Checked no regression in rendering the player in the world

image

You now can see normals are applied well in the inventory screen too.

image

image

With a shader turned on, I also found no issues with this fix.

_For reminder: this is not an issue in 1.20.1._

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed the target indicator invisible for non Epic Fight patched entities
 - Fixed the lock-on target not being synced to the server when using mouse snap to change the target
+- Fixed the player being dark in inventory screen
 
 ### Changed
 
--- a/src/main/java/yesman/epicfight/api/utils/math/MathUtils.java
+++ b/src/main/java/yesman/epicfight/api/utils/math/MathUtils.java
@@ -288,11 +288,13 @@ public class MathUtils {
 		poseStack.scale(vector.x(), vector.y(), vector.z());
 	}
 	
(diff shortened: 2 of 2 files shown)


EchoEllet (approved):
Thanks for the detailed explanation. 
LGTM.

PR #2318, opened 2025-12-10
[1.20.1] Fixed datapack animation loading failure

Yesssssman:
Caused by a simple mistake, "/" was accidentally omitted in the conversion method, leading `NoSuchElementException`. (fixes #2312)

image

_Demonstration:_

https://github.com/user-attachments/assets/b6913dff-6f00-4029-b65c-4fc9a14341fd

Bokken, our official sample datapack, works fine.

PR #2319, opened 2025-12-10
[1.21.1] Fixed datapack animation loading failure

Yesssssman:
Same with #2318, but 1.21.1.

Note: We do not have a sample datapack that meets the 1.21.1 form, so we can't test whether this works correctly. Tho I wanted to fix what was an apparent malfunction.


EchoEllet (approved):
LGTM.

PR #2323, opened 2025-12-10
Fix issue regarding air attack and basic attack not activating even if there is no stamina consumption.

Forixaim:
This PR fixes the issue regarding the air attack being executed whether the player is off ground or not. It also fixes the issue regarding basic attack not executing.


src/main/java/yesman/epicfight/world/capabilities/item/CapabilityItem.java
```diff
@@ -352,7 +352,12 @@ public boolean availableOnHorse() {
 	}
 	
 	public boolean checkOffhandValid(LivingEntityPatch<?> entitypatch) {
-		return this.getStyle(entitypatch).canUseOffhand() && EpicFightCapabilities.getItemStackCapability(entitypatch.getOriginal().getOffhandItem()).canHoldInOffhandAlone();
+        try {
+            return this.getStyle(entitypatch).canUseOffhand() && EpicFightCapabilities.getItemStackCapability(entitypatch.getOriginal().getOffhandItem()).canHoldInOffhandAlone();
+        }
+        catch (NullPointerException e) {
+            return false;
+        }
```
EchoEllet, line 360: Is catching `NullPointerException` a good approach here?

This exception is a programming error (an indication of a bug) and not a runtime user-facing failure, so it should not be handled.

Instead, maybe we should consider checking for `null` and preventing it from happening in the first place?

src/main/java/yesman/epicfight/main/EpicFightMod.java
```diff
@@ -5,6 +5,7 @@
 import java.util.function.BiConsumer;
 import java.util.function.Function;
 
+import com.mojang.logging.LogUtils;
```
EchoEllet, line 8: Shouldn't we use `EpicFightMod.LOGGER`? 
This will keep it consistent with most mods, allowing us to determine which log comes from which mod.

src/main/java/yesman/epicfight/client/events/ClientEvents.java
```diff
@@ -173,6 +175,13 @@ public static void clientRespawnEvent(ClientPlayerNetworkEvent.Clone event) {
 		ClientEngine.getInstance().controlEngine.setPlayerPatch(newCap);
 		ClientEngine.getInstance().renderEngine.initHUD();
 	}
+
+    @SubscribeEvent(priority = EventPriority.HIGHEST)
+    public static void onClientTick(TickEvent.ClientTickEvent event) {
+        if (MINECRAFT.getOverlay() == null && MINECRAFT.screen == null) {
+            ClientEngine.getInstance().controlEngine.handleEpicFightKeyMappings();
+        }
+    }
```
EchoEllet: I think Yesman's PR added a client tick event to allow us to move away from these Forge APIs but few notes:

- This will run twice (two phases, start and end), and you should aim to call this only at the end
  - In NeoForge, this is just `ClientTickEvent.Post` instead of `TickEvent.ClientTickEvent`
  - In Forge, you should use `event` and return early if `event.phase` is `START`
- Is there a reason why we're using `priority = EventPriority.HIGHEST`? If yes, we should document it for future readers (including us).
- For @Yesssssman is your branch ready? Just want to ensure new code is multi-platform friendly and independent and fully decoupled of MinecraftForge

EchoEllet (changes requested):
Looks good, with a few notes.

src/main/java/yesman/epicfight/world/capabilities/item/CapabilityItem.java
EchoEllet: [Source](https://docs.oracle.com/javase/8/docs/api/java/lang/NullPointerException.html)

<details><summary>Images</summary>
<p>

image


image

image

</p>
</details>

src/main/java/yesman/epicfight/main/EpicFightMod.java
Forixaim: yeah we should, this is part of the old legacy stuff from the previous branch.

src/main/java/yesman/epicfight/client/events/ClientEvents.java
Forixaim: removed.

src/main/java/yesman/epicfight/client/events/ClientEvents.java
```diff
@@ -12,7 +12,9 @@
 import net.minecraftforge.api.distmarker.Dist;
 import net.minecraftforge.client.event.ClientPlayerNetworkEvent;
 import net.minecraftforge.client.event.ScreenEvent;
+import net.minecraftforge.event.TickEvent;
 import net.minecraftforge.event.entity.player.PlayerInteractEvent;
+import net.minecraftforge.eventbus.api.EventPriority;
```
EchoEllet, line 17: If this is not used anymore, shouldn't we remove it?

We're trying to be less dependent on MinecraftForge or third-parties for a more future-proof Epic Fight.

src/main/java/yesman/epicfight/client/events/ClientEvents.java
Yesssssman: I'm working on UI refactor stuff first trying to end up emote feature in this year. Event refactoring is not on the first place atm

Forixaim:
Closing PR to keep itself up to date.

PR #2336, opened 2025-12-14
TPS Camera and RightClickEvent

OneWorldStudio:
You had an error with the TPS camera, but I fixed it and it's working properly now
I just needed version 20.14, but I couldn't use it, so I decided to contribute.

image
image

https://youtu.be/rLt7aMZpQZY
https://youtu.be/vmifEkaw-mg


guivnf:
Why was this PR closed? We were still reviewing it 😅

OneWorldStudio:
> Why was this PR closed? We were still reviewing it 😅

I closed it because I found another problem with the Connector, and the error will remain with this fix

OneWorldStudio:
I will offer a different option.
A little later

PR #2337, opened 2025-12-14
[1.21.1] Prepare for ResourceLocation rename and add more DRY static utility method

EchoEllet:
This change is backward compatible (deprecate methods instead of removing them).

Similar to https://github.com/EchoEllet/dragonfist-legacy/pull/2 but for Epic Fight, to make the migration to 1.21.11 easier.

[Mojang renamed `ResourceLocation` to `Identifier` in 1.21.11](https://neoforged.net/news/21.11release/#renaming-of-resourcelocation-to-identifier).

To handle the change in 1.21.11:

```diff
- ResourceLocation key = BuiltInRegistries.ITEM.getKey(...);
+ Identifier key = BuiltInRegistries.ITEM.getKey(...);

- register(ResourceLocation.fromNamespaceAndPath("mymod", "thing"), ...);
+ register(Identifier.fromNamespaceAndPath("mymod", "thing"), ...);
```

This PR makes it easier by avoiding `ResourceLocation.fromNamespaceAndPath(EpicFightMod.MOD_ID, $path)`, and replacing it with `EpicFightMod.identifier($path)`.

Except for the `EpicFightMod` class, most of the changes were automated.

image

### DIFF
--- a/src/main/java/yesman/epicfight/api/animation/types/DynamicAnimation.java
+++ b/src/main/java/yesman/epicfight/api/animation/types/DynamicAnimation.java
@@ -142,7 +142,7 @@ public abstract class DynamicAnimation {
 	}
 	
 	public ResourceLocation getRegistryName() {
-		return ResourceLocation.fromNamespaceAndPath(EpicFightMod.MODID, "");
+		return EpicFightMod.identifier("");
 	}
 	
 	public int getId() {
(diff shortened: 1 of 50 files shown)


src/main/java/yesman/epicfight/api/animation/types/DynamicAnimation.java
```diff
@@ -142,7 +142,7 @@ public boolean canBePlayedReverse() {
 	}
 	
 	public ResourceLocation getRegistryName() {
-		return ResourceLocation.fromNamespaceAndPath(EpicFightMod.MODID, "");
```
EchoEllet, line 145: QQ: Why was this an empty String? It might be worth documenting the reason (if there is any).

PR #2345, opened 2025-12-19
TPS Camera and RightClickEvent (RenderEngine)

OneWorldStudio:
You had an error with the TPS camera, but I fixed it and it's working properly now
I just needed version 20.14, but I couldn't use i

https://youtu.be/rLt7aMZpQZY
https://youtu.be/vmifEkaw-mg


src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
```diff
@@ -131,7 +131,7 @@ private EpicFightCameraAPI() {
 	 * When zooming ranged weapons or TPS mode is turned on by config
 	 */
 	public boolean isTPSMode() {
-		if (this.minecraft.options.getCameraType() == CameraType.THIRD_PERSON_BACK && ClientConfig.cameraMode.shouldSwitch(this)) {
+		if (this.minecraft.options.getCameraType() == CameraType.THIRD_PERSON_BACK && ClientConfig.cameraMode != null && ClientConfig.cameraMode.shouldSwitch(this)) {
```
Yesssssman, line 134: What's the cause of making `ClientConfig.cameraMode` as null? Does the null check here necessary?

src/main/java/yesman/epicfight/client/events/engine/RenderEngine.java
```diff
@@ -792,11 +792,26 @@ public static void rightClickBlockEvent(PlayerInteractEvent.RightClickBlock even
...
+			EpicFightCapabilities.getUnparameterizedEntityPatch(event.getEntity(), LocalPlayerPatch.class).ifPresent(playerpatch -> {
 				Vec3 toHit = event.getHitVec().getLocation().subtract(playerpatch.getOriginal().getEyePosition());
-				playerpatch.getOriginal().setXRot((float)MathUtils.getXRotOfVector(toHit));
-				playerpatch.getOriginal().setYRot((float)MathUtils.getYRotOfVector(toHit));
+
+				float xRot = (float)MathUtils.getXRotOfVector(toHit);
+				float yRot = (float)MathUtils.getYRotOfVector(toHit);
+
+				playerpatch.getOriginal().setXRot(xRot);
+				playerpatch.getOriginal().xRotO = xRot;
```
Yesssssman, line 813: Is there any specific reason to sync the old rotations? If we correct the old rotations too it makes the player instantly looking at the destination without a short interpolation and it looks blazing fast.

```java
    Vec3 toHit = event.getHitVec().getLocation().subtract(playerpatch.getOriginal().getEyePosition());
	float xRot = (float)MathUtils.getXRotOfVector(toHit);
	float yRot = (float)MathUtils.getYRotOfVector(toHit);

	playerpatch.getOriginal().setXRot(xRot);
	playerpatch.getOriginal().xRotO = xRot;

	playerpatch.getOriginal().setYRot(yRot);
	playerpatch.getOriginal().yRotO = yRot;

	playerpatch.getOriginal().setYHeadRot(yRot);
	playerpatch.getOriginal().yHeadRotO = yRot;
```
This code makes:

https://github.com/user-attachments/assets/28e70765-e200-4f9b-8dff-89bf7c1d8417

which makes the player look at the lever in time, but

```java
    Vec3 toHit = event.getHitVec().getLocation().subtract(playerpatch.getOriginal().getEyePosition());
	float xRot = (float)MathUtils.getXRotOfVector(toHit);
	float yRot = (float)MathUtils.getYRotOfVector(toHit);

	playerpatch.getOriginal().setXRot(xRot);
	playerpatch.getOriginal().setYRot(yRot);
	playerpatch.getOriginal().setYHeadRot(yRot);
```

This code gives a smooth transition so it looks more natural

https://github.com/user-attachments/assets/f892a56b-32c1-4210-85a9-8053807e644f

OneWorldStudio:
image

image

Yesssssman (changes requested):

Yesssssman:
@OneWorldStudio Sorry, I forgot to submit the review. Please take a look at my suggestions.

src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
OneWorldStudio: for some strange reason without this I could not run the mod Shoulder Surfing Reloaded and Shoulder Surfing Reloaded x Iron's Spells

This may not be the best solution I could think of

src/main/java/yesman/epicfight/client/events/engine/RenderEngine.java
OneWorldStudio: I agree, I didn't think about it.

OneWorldStudio:
> @OneWorldStudio Sorry, I forgot to submit the review. Please take a look at my suggestions.

Yes, everything is fine, thank you for taking the time to reply.

src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
EchoEllet: > This may not be the best solution I could think of

Actually, there is a WIP branch](https://github.com/Epic-Fight/epicfight/tree/feat/iron-spells-camera-support) for adding camera compatibility with Iron's Spells and Spellbooks.

We're still waiting for Iron's Spells to provide a stable public API over mixing internals (https://github.com/iron431/irons-spells-n-spellbooks/pull/960).

Yesssssman: > for some strange reason without this I could not run the mod Shoulder Surfing Reloaded and Shoulder Surfing Reloaded x Iron's Spells
> 
> This may not be the best solution I could think of

I think it's unrelated to the issue? cuz checking ClientConfig.cameraMode is `null` really does nothing because it can't be `null` at all. (Tho there could be some unknown issue)

Yesssssman:
I'll just merge the PR and will manually fix what I pointed out, thanks for your effort!

src/main/java/yesman/epicfight/api/client/camera/EpicFightCameraAPI.java
OneWorldStudio: you're right, it really doesn't work, it might have been a one-time bug with me

EchoEllet: This issue has been addressed by using en Epic Fight event over changing the client confif when Shoulder Surfing Reloaded is installed.

EchoEllet: This issue has been addressed by using en Epic Fight event over changing the client confif when Shoulder Surfing Reloaded is installed.

Issue #2332, opened 2025-12-13
Provide Custom Maven instead of depending on CurseForge/Modrinth repositories


EchoEllet:
**Most mods already provide this:** including Iron's Spells and spellbooks, GeckoLib, AzureLib, Curios API, JEI, playerAnimator, ModMenu, Controlify, Zoomify, YACL, Kotlin for Forge, Fabric API, KubeJS, Architectury API, Cloth Config, Rhino, Caelus, and most well-known community mods.

I believe it will enhance the developer experience:

- Transitive dependencies, metadata, sources, JAR files, and documentation are built-in without the need for manual download for each version
- Readable version Strings
  - Also allow the addon to display the tested EF version in release notes ([example](https://modrinth.com/mod/epic-fight-invincible-lib/version/21.14.6.3-mc1.21.1-neoforge)).
- Support for Gradle modules or separate JAR files (`epicfight-common`, `epicfight-neoforge`, `epicfight-api`, and so on)
- Allow to publish snapshots or dev versions or early versions to give addons time to migrate without publishing the version directly to CurseForge or Modrinth, and without manually building JAR files and then asking them to manually change in `libs` and then change back to the one from CurseForge/Modrinth (inconvenient)
- CurseForge or Modrinth is blocked in many countries, and there is no strong reason to force the use of a VPN when we can fix it during development
- CurseForge and Modrinth maven repositories are not reliable; sometimes they stop working or block third-party distribution
  - It's also more professional and shows that we care about maintenance quality

Consumers can still use any Maven they want, and we can keep this fully automated, so no extra chores. I think we should do it at some point, but we need to agree on a plan now.

<img width="1596" height="968" alt="Image" src="https://github.com/user-attachments/assets/156e1bcd-c09a-4235-92d1-57e1ee5e8d79" />

EchoEllet:
This is important for multi-platform support, to allow third-parties depending on `common`, `neoforge`, and `fabric` modules.

The most important part is that we can view the Epic Fight source code directly in the IDE, use readable version strings, and being able to compile aginst public API only.

Yesssssman:
I actually wanted to make Epic Fight own repository in maven but I'm very unfamiliar with that. This is the stance we will have to take in the future.

EchoEllet:
> I actually wanted to make Epic Fight own repository in maven but I'm very unfamiliar with that. This is the stance we will have to take in the future.

[...]

PR #2350, opened 2025-12-23
[1.21.1] Support Automatic Modrinth Sources JAR download

EchoEllet:
Changes the JAR file name from `epicfight` -> `epic-fight`, and therefore matches the Modrinth project [slug URL to enable automatic sources JAR file](https://support.modrinth.com/en/articles/8801191-modrinth-maven#h_1b24106498) for convenient development.

CurseForge Maven doesn't support this feature, so the JAR file name format is irrelevant. 

See also [this Epic Fight WIKI](https://epicfight-docs.readthedocs.io/API/Starting/#setting-up-your-gradle-build).

The ideal way is to address #2332, but for now, this is good enough.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -48,7 +48,11 @@ repositories {
 }
 
 base {
-    archivesName = modId
+    // "epic-fight" is intentionally being used instead of the mod ID,
+    // to keep the JAR file name consistent with the project slug URL,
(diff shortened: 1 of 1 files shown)


sentwayfarer (approved):
LGTM

Yesssssman (approved):
Looks valid reason

PR #2351, opened 2025-12-23
[1.21.1] Add missing @NotNull annotations

EchoEllet:
Adds missing `@NotNull` annotations for parameters that are not intended to accept a `null` argument.


Yesssssman (approved):
LGTM

EchoEllet:
**FYI:** This is not a breaking change for Java consumers, but is for Kotlin addons, since the Kotlin compiler makes use of this annotation to enforce it at compile-time.

Kotlin assumes it's nullable by default (without the annotation) when calling Java code, so it's `Entity?`, and with the annotation it becomes `Entity`.

However, it's a good breaking change because it's better to handle programming errors at compile time rather than at runtime.

Yesssssman:
Nice decision that we'd keep Java for our main language

EchoEllet:
> Nice decision that we'd keep Java for our main language

FYI: It's not a bug, it's as intended, and good interoperability with Java.

PR #2360, opened 2025-12-30
Set up Gradle Projects for multi-platform foundation (common, fabric, neoforge)

EchoEllet:
Setups `common`, `neoforge`, and `fabric` Gradle projects (AKA modules) for multi-platform progress, without using third-party templates, plugins, or dependencies (i.e., **without Architectury API or Gradle Plugin**).

The source code has been moved fully from the root project `src/` to `neoforge/src/`, while the vanilla assets (including `assets/epicfight`) have been moved to `common` but some NeoForge exclusive assets, such as `data/neoforge` kept under `neoforge`, for clear separation of concerns, so they are not included in the Fabric JAR file.

> [!TIP]
> This is **100% backward compatible with existing addons**, since it does not affect how addons import the Java classes.
> A single Gradle project may contain multiple packages and resources. All packages and resources under `common` are included in `neoforge` and `fabric`, but not vice versa.

> [!NOTE]
> This was done using standard Gradle projects, promoting best practices whenever possible, such as using Gradle convention plugins (`gradle/build-logic`).
> `gradle/build-logic` (precompiled plugins) is newer than `buildSrc`, for faster builds, and is officially recommended by Gradle, while supporting Gradle 9, configuration-cache, and up-to-date tools.

<details><summary>The Fabric version of Epic Fight is empty.</summary>
<p>

Can be run via `./gradlew fabric:runClient`.

image

</p>
</details> 

> [!IMPORTANT]
> It's important to use `./gradlew neoforge:publishMods` and not `./gradlew publishMods`, otherwise the empty Fabric version will be published to Modrinth and CurseForge, which is only appropriate in April.

<details><summary>Publish the NeoForge version using the Gradle plugin via the IntelliJ IDEA GUI</summary>
<p>

image

</p>
</details> 

<details><summary>Screenshots while crafting</summary>
<p>

image

</p>
</details> 

<details><summary>NeoForge Vanilla-mode</summary>
<p>

NeoForge provides [NeoForm](https://github.com/neoforged/NeoForm) ([vanilla mode](https://github.com/neoforged/ModDevGradle#vanilla-mode)), which allows using the NeoForge toolchain in the `common` project, but vanilla Minecraft classes, without any patches or NeoForge APIs.

image

This is used in the `common` project, so even though there is the `neoforge {}` block, it's not NeoForge, and this was made for compatibility reasons.
This approach is friendly with other mod-platform such as Fabric.

</p>
</details> 

### Motivation

<details><summary>Why was this done?</summary>
<p>

We can't progress without this Gradle setup, since it's more than just abstracting `net.neoforged.*`.

Forge-like platforms patch vanilla classes to add new fields, methods, and constructors.

> **Note:** NeoForge is improving the situation in MC 1.21.9+ and 26+.

The same vanilla class may not work outside of Forge and will be silently dependent on Forge

image

image

For example, `Mob#spawnType` and `Mob#spawnCancelled` were added by Forge, and it's hard to tell whether a specific MC API was added by Forge or not without looking at this directory: https://github.com/neoforged/NeoForge/tree/1.21.x/patches/net/minecraft (NeoForge patches).

The ideal solution is to set up Gradle projects (AKA modules), which are `common`, `fabric`, and `neoforge`,
where the Java source code is fully in `neoforge` (`src/main` -> `neoforge/src/main`).

The situation can be improved by slowly abstracting and moving classes to `common` with the corresponding `neoforge` and `fabric` platform implementations to catch early
design issues.

This is not **a code-breaking change**; for example, if a class was moved from `neoforge` to `common` without any source modification, addons can still import the same class with no behavior changes.

</p>
</details> 

<details><summary>Note regarding existing and new code</summary>
<p>

1. New features should be in `common`, designed in mind to be independent of a specific mod-loader.
2. Existing features can be slowly moved from `neoforge` to `common`, and then implement platform-specific methods in `fabric`.

</p>
</details> 

<details><summary>Isn't this maintenance burden?</summary>
<p>

A multi-platform architecture does not exists primarly to support Fabric, but to be independent of Forge or a specific mod-loader, and therefore a more future-proof project.

This will be significantly useful if:

1. Minecraft introduced official modding APIs.
2. Forge/NeoForge got forked again.
3. The maintainers decided to move to another mod-loader for whatever reason.
4. NeoForge introduced public API breaking changes, which were done multiple times to clean legacy decisions, so a multi-platform setup will be very useful for porting to newer MC versions.

This was already discussed quite a few times in Discord and internally with the Epic Fight team.

</p>
</details> 

<details><summary>Why not StoneCutter?</summary>
<p>

Gradle projects don't change the Java packages or how they are imported. It's just a powerful way to separate builds, scripts, and dependencies.

StoneCutter, while it's great for multi-version, it's heavily dependent on preprocessor comments:

image

So it's less modular, as all platform implementations are in the same Gradle project (`src/main`), and a specific platform cannot be removed easily by simply removing the corresponding project directory.

It's harder to share the platform JAR files for third parties, and therefore, it's less efficient for other multi-platform mods to integrate with Epic Fight, unless they use StoneCutter too.

StoneCutter can still be used for MC multi-version, but the setup is complex, and the build performance is not ideal for convenience.

Also, it does not solve the issue of third-party dependencies.

</p>
</details> 

### Non-goals

- Continue supporting Forge, or explicit Quilt support.
- Complete Fabric port in `1.20.1` or `1.21.1`.
- Supporting alternative mod platforms, such as Liteloader.
  - This step will make it easier for third parties to integrate; however, it's not officially planned or something we are looking forward to, due to the amount of work and considerations required.
- Discounting the NeoForge support.

### Other Questions

<details><summary>What is libs.versions.toml?</summary>
<p>

The versions are stored in Gradle version catalogs (`libs.versions.toml`) instead of `gradle.properties` (the legacy approach), which is more suitable for multi-project sharing of Gradle coordinates, versions, and plugins in a type-safe convivnent way.

image

This also makes the `dependencies` in `common`, `neoforge`, and `fabric` more readable and minimal, allowing for changes to versions, Maven repositories, or coordinates from one central place.

</p>
</details> 

<details><summary>What is strictMaven() function?</summary>
<p>

Preferred over the regular `maven { ... }` as it filters to specific groups and not all of them to reduce network calls, and therefore faster builds.
Typically, mod Maven repositories include several groups and dependencies that are not needed, so it's better not to look them up.

```diff
- maven {
-    name = "NeoForged"
-    url = uri("https://maven.neoforged.net/releases")
- }
+ strictMaven(
+    name = "NeoForged",
+    url = "https://maven.neoforged.net/releases",
+    includeGroups = "net.neoforged",
+ )
```

</p>
</details>


Forixaim (approved):
LGTM.

EchoEllet:
<details><summary>Complete Fabric support may arrive in MC 26 or newer versions</summary>
<p>

Due to the major changes in Minecraft 26, `1.21.11` and `1.21.9`, it's more future-proof to discount `1.21.1` and `1.20.1` to move on to 26, and therefore the multi-platform progress will be in the newer branches.

image

However, this step can still be done regardless of whether Fabric is supported or not in the current version, to minimize Git conflicts and allow progress in newer/existing feature branches.

</p>
</details>

fabric/src/main/resources/fabric.mod.json
sentwayfarer, line 1: It seems that the `authorsList` doesn't have quotes around it like the rest of them. Syntax error?

sentwayfarer (changes requested):
One change needed.

fabric/src/main/resources/fabric.mod.json
EchoEllet: Good question. 

This will be replaced with a valid JSON list by Gradle scripts by processing the`authors` field in `gradle.properties` (names separated by a comma).

```kotlin
// gradle/build-logic/conventions/src/main/kotlin/SharedUtils.kt
"authorsList" to modAuthors
            .split(",").joinToString(
                prefix = "[\"",
                postfix = "\"]",
                separator = "\", \""
            ) { it.trim() },
```

The IDE or any tooling thinks this is an invalid JSON file, but it's totally valid.

Run `./gradlew fabric:build` and then take a look at the processed JSON file in `fabric/build/resources/main/fabric.mod.json`.

That's the change:

```diff
- "authors": ${authorsList},
+ "authors": ["Yesman(a.k.a maninthe_home)", "Gui", "Asan", "Squoshi", "Fori", "Ellet"],
```

Yes, I agree that we need to do something about the fact that the IDE and other tools think it's invalid JSON, and it's confusing, but the only workaround I see is:

```json
"authors": [
    "${authors}"
  ],
```

<details><summary>Details</summary>
<p>

When all authors are in the same `String`:

image

When each author is in a separate `String` in the JSON list:

image




</p>
</details> 

Maybe this is better:

```json
"authors": [ ${authorsList} ],
```

EchoEllet: > Maybe this is better:

I have [changed it in this push](https://github.com/Epic-Fight/epicfight/compare/8acd12e97e7136d9011135c195ed69ebcfb485e6..9dfc5e663d359c516a8617a55fd511525c9ed20f) with a clearer name, and now the `[]` is part of the source JSON file.

sentwayfarer (approved):
LGTM

EchoEllet:
@Yesssssman
I have rebased the branch to address all conflicts with the recent changes.

<details><summary>The CI failure is irrelevant to the changes</summary>
<p>

image

It seems that there is an unchecked file that is used in a checked file.

</p>
</details>

Yesssssman (approved):
Looks cool, still I'm trying to understand multi loader shape by myself through local dummy project. That way I can figure out some details of this structure.

PR #2365, opened 2026-01-03
Initial ModPlatform and ClientModPlatform design

EchoEllet:
Introduces `ModPlatform`, with two methods, `isDevelopmentEnvironment` and `isModLoaded`, as an example of what this interface does.

It shouldn't provide every mod-loader method directly in here; for example, we could introduce an interface `DeferredRegistrar` for anything related to deferred registration with all methods in there, and `ModPlatform` only maintains the instance of `DeferredRegistrar`. This design forces the platform modules to provide `DeferredRegistrar`.

This also introduces the `EpicFight` class, which is the common code, and adds TODOs for renaming `EpicFightMod` in `neoforge` to `EpicFightNeoForge`. That will be done in the next PR for an easier review.

`neoforge` and `fabric` must call `EpicFight#initialize()`, which calls `ModPlatformProvider.initialize()`, which requires a `ModPlatform` argument.

NeoForge provides `NeoForgeModPlatform`, and `fabric` provides `FabricModPlatform`.

### Design

**ModPlatform:**

```java
/// Provides access to mod-loader specific functionalities that cannot be achieved using vanilla classes otherwise.
public interface ModPlatform {
    boolean isDevelopmentEnvironment();

    boolean isModLoaded(@NotNull final String id);
}

public final class NeoForgeModPlatform implements ModPlatform {
    @Override
    public boolean isDevelopmentEnvironment() {
        return !FMLLoader.isProduction();
    }

    @Override
    public boolean isModLoaded(@NotNull final String id) {
        return ModList.get().isLoaded(id);
    }
}

public final class FabricModPlatform implements ModPlatform {
    @Override
    public boolean isDevelopmentEnvironment() {
        return FabricLoader.getInstance().isDevelopmentEnvironment();
    }

    @Override
    public boolean isModLoaded(@NotNull final String id) {
        return FabricLoader.getInstance().isModLoaded(id);
    }
}
```

**Common vanilla entrypoint:**

```java
/// Common functionalities shared between the platform entrypoints (EpicFightNeoForge and EpicFightFabric).
public final class EpicFight {
    private EpicFight() {
    }

    public static final String MODID = "epicfight";
    public static final Logger LOGGER = LogManager.getLogger(MODID);


    /// Initializes common functionalities.
    ///
    /// Provides the global [ModPlatform] instance given by the mod platform (e.g., NeoForge, Fabric)
    /// to common vanilla code for sharing.
    public static void initialize(@NotNull final ModPlatform platform) {
        ModPlatformProvider.initialize(platform);
    }
}
```

**Fabric platform entrypoint:**

```java
public final class EpicFightFabric implements ModInitializer {
    @Override
    public void onInitialize() {
        EpicFight.initialize(new FabricModPlatform());
    }
}
```

**NeoForge platform entrypoint:**

```java
// FYI: EpicFightNeoForge is currently named EpicFightMod, and there is a TODO to track renaming it in the next PR
@Mod(EpicFight.MODID)
public final class EpicFightNeoForge {
    public EpicFightNeoForge() {
        EpicFight.initialize(new NeoForgeModPlatform());
    }
}
```

**Usage in common vanilla code:**

```java
if (ModPlatformProvider.get().isModLoaded("epicskills")) {}
```

The same pattern applies with `ClientModPlatform` and `ClientModPlatformProvider`, and technically, client-only methods can be merged directly into `ModPlatform`, but this is not the recommended vanilla approach. Both NeoForge and Fabric encourage avoiding importing client-only classes in common side classes through a clear separation.

The stronger separation will blend well with [the upcoming NeoForge source sets separation](https://neoforged.net/news/21.5release/#split-sourcesets).

### DIFF
--- a/common/src/main/java/yesman/epicfight/platform/ModPlatform.java
+++ b/common/src/main/java/yesman/epicfight/platform/ModPlatform.java
@@ -1,0 +1,10 @@
+package yesman.epicfight.platform;
+
+import org.jetbrains.annotations.NotNull;
+
+/// Provides access to mod-loader specific functionalities that cannot be achieved using vanilla classes otherwise.
+public interface ModPlatform {
+    boolean isDevelopmentEnvironment();
+
+    boolean isModLoaded(@NotNull final String id);
+}
--- a/common/src/main/java/yesman/epicfight/EpicFight.java
+++ b/common/src/main/java/yesman/epicfight/EpicFight.java
@@ -1,0 +1,33 @@
+package yesman.epicfight;
+
+import net.minecraft.resources.ResourceLocation;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.jetbrains.annotations.NotNull;
+import yesman.epicfight.platform.ModPlatform;
+import yesman.epicfight.platform.ModPlatformProvider;
+
+/// Common functionalities shared between the platform entrypoints.
+public final class EpicFight {
+    private EpicFight() {
+    }
+
+    public static final String MODID = "epicfight";
+    public static final String EPICSKINS_MODID = "epicskins";
+    public static final Logger LOGGER = LogManager.getLogger(MODID);
+
+    /// Creates an identifier that points to an Epic Fight resource.
+    ///
+    /// This was called `identifier` and not `resourceLocation` since [Mojang renamed `ResourceLocation` to `Identifier` in 1.21.11](https://neoforged.net/news/21.11release/#renaming-of-resourcelocation-to-identifier).
+    public static @NotNull ResourceLocation identifier(@NotNull final String path) {
+        return ResourceLocation.fromNamespaceAndPath(MODID, path);
+    }
+
+    /// Initializes common functionalities.
+    ///
+    /// Provides the global [ModPlatform] instance given by the mod platform (e.g., NeoForge, Fabric)
+    /// to common vanilla code for sharing.
+    public static void initialize(@NotNull final ModPlatform platform) {
+        ModPlatformProvider.initialize(platform);
+    }
+}
--- a/common/src/main/java/yesman/epicfight/EpicFightClient.java
+++ b/common/src/main/java/yesman/epicfight/EpicFightClient.java
@@ -1,0 +1,20 @@
+package yesman.epicfight;
+
+import org.jetbrains.annotations.NotNull;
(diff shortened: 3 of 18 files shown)


EchoEllet:
This has already discussed in DMs with @Yesssssman (although we should use GitHub issues for future PRs, with the rest of the team).

common/src/main/java/yesman/epicfight/platform/ModPlatform.java
```diff
@@ -0,0 +1,8 @@
+package yesman.epicfight.platform;
+
+/// Provides access to mod-loader specific functionalities that cannot be achieved using vanilla classes otherwise.
+public interface ModPlatform {
+    boolean isDevelopmentEnvironment();
+
+    boolean isModLoaded(String id);
```
Yesssssman: It will be great if we can have something like `<T> net.minecraft.core.Registry<T> getRegistry(ResourceKey<T> key);` so we don't have to implement each registry's getter (as we discussed in DM)

Yesssssman (approved):

sentwayfarer (approved):
LGTM

common/src/main/java/yesman/epicfight/platform/ModPlatform.java
EchoEllet: I'm not sure if this can work universally in a scalable way. Forge, for example, has custom `ForgeRegistries`, so the approach of using one method with vanilla types may not work on Forge.

The goal isn't to support Forge, but a design that works well, and we only have 10 registries, so we can provide explicit methods such as `registerItem`, `registerEntity`, `registerSound`, and `registerCreativeModeTab`.

This way, we have full flexibility and different return types and parameters:

```java
import java.util.function.Supplier;

public interface DeferredRegistrar {
    
    <T extends Item> DeferredItem<T> registerItem(String name, Supplier<T> builder);

    <T extends Entity> DeferredEntity<T> registerEntity(
        String name, 
        EntityType.Builder<T> builder
    );

    DeferredSoundEvent registerSound(String name);

    DeferredCreativeModeTab registerCreativeModeTab(
        String name, 
        Supplier<CreativeModeTab> builder
    );
}
```

This way, we can use `DeferredRegister.Items` provided by NeoForge for `registerItem`, which is a more specific type than `DeferredRegister<Item>`.

Each method is usually one line, so this won't result in bloated implementations.

image

PR #2352, opened 2025-12-24
[1.21.1] Type Safe Entity Patch Registration

EchoEllet:
There was an error in `1.21.1` (#2329), which causes the vindicator Epic Fight patch to be fully broken (spams the log as well). It was due to casting `Vindicator` as `Spider`.
This PR enforces a cleaner design that is less error-prone, easier, and more convenient to use with documentation.

**Before:**

```java
registry.put(EntityType.PLAYER, entity -> new ServerPlayerPatch((ServerPlayer)entity));
registry.put(EntityType.ZOMBIE, entity -> new ZombiePatch<> ((Zombie)entity));
registry.put(EntityType.CREEPER, entity -> new CreeperPatch((Creeper)entity));
registry.put(EntityType.ARROW, entity -> new ArrowPatch((Arrow)entity));
registry.put(EntityType.SPECTRAL_ARROW, entity -> new ArrowPatch((SpectralArrow)entity));
```

**After:**

```java
registerEntityPatchUnsafe(registry, EntityType.PLAYER, entity -> new ServerPlayerPatch((ServerPlayer) entity));
registerEntityPatch(registry, EntityType.ZOMBIE, ZombiePatch::new);
registerEntityPatch(registry, EntityType.CREEPER, CreeperPatch::new);
registerEntityPatch(registry, EntityType.ARROW, ArrowPatch::new);
registerEntityPatch(registry, EntityType.SPECTRAL_ARROW, ArrowPatch::new);
```

The only case where `registerEntityPatchUnsafe` is needed is when dealing with some vanilla types, which is the case with `ServerPlayerPatch`.

This improvement applies to addons as well, since putting the patch directly to the `Map` is not type-safe and is not developer-friendly.

```diff
public void registerEntityPatches(EntityPatchRegistryEvent event) {
-    event.getTypeEntry().put(ModEntityTypes.EXAMPLE, entity -> new ExamplePatch<>((Example)entity));
+    event.registerEntityPatch(ModEntityTypes.EXAMPLE, ExamplePatch::new);
}
```

I already have these utilities in my [Epic Fight addon](https://github.com/EchoEllet/dragonfist-legacy), though I think it makes sense to merge them to the upstream core mod.

image

<details><summary>Tested the dedicated server with arrows</summary>
<p>

image

</p>
</details> 

<details><summary>A screenshot while cleaning up</summary>
<p>

image

</p>
</details>

PR #2368, opened 2026-01-03
Deprecate EntityPatchRegistryEvent#getTypeEntry()

EchoEllet:
Deprecates `EntityPatchRegistryEvent#getTypeEntry()`, since `#registerEntityPatch` was introduced in #2352.


EchoEllet:
I already discussed this with @Yesssssman, and it doesn't seem like there is any use case in keeping accessing the `getTypeEntry()` method public. However, it's hard to know, so it will be deprecated, and we will keep contacting the addon devs to see if it's used and cannot be replaced with `registerEntityPatch` / `registerEntityPatchUnsafe`.

Forixaim (approved):

Yesssssman (approved):
LGTM

PR #2373, opened 2026-01-04
Migrate MODID, LOGGER, identifier fields and properties to common EpicFight

EchoEllet:
- Renames `EpicFightMod` to `EpicFightNeoForge` since it's inside the `neoforge` Gradle project.
  - A deprecated `EpicFightMod` class has been introduced to avoid breaking addons and serve as docs.
- Uses `LOGGER`, `MODID`, `identifier()`, and `prefix()` from `EpicFight` (in `common`) over `EpicFightNeoForge`.
  -  Automated via IDEA structural replacement.
  -  The old methods in `EpicFightNeoForge` are deprecated and will be removed in the future. It's kept to serve as documentation for current addon devs, while keeping it **backward compatible**.
- Clears two TODOs that were added in #2365

This allows sharing code between `fabric` and `neoforge`, and keeping the code more independent of Forge-like platforms without any breaking changes.


EchoEllet:
The changes are automated, so it might be worth doing that later once @Yesssssman finishes porting to `1.21.11` to minimize the Git conflicts.

PR #2383, opened 2026-01-05
Add required and optional dependencies when publishing the mod

EchoEllet:
Adds the optional and required mod dependencies.
For now, NeoForge doesn't have any required mod dependencies. That may change if [Player Animation Library](https://modrinth.com/mod/player-animation-library) is used as an animation library for Epic Fight.


Yesssssman (approved):
LGTM

PR #2386, opened 2026-01-06
Allow building the project with "Unreleased" mark

Yesssssman:
This is a problem when we want to share a build internally for the features that are incomplete. The mod publishing plugin searches for proper sentences in `CHANGELOG.md` by [version] - yyyy-mm-dd format but private builds doesn't have release date.

I modified regular expression that searches the changelog content from .md file so it can detect `Unreleased` mark instead of the date format. In that way we can build the project without modifying a random date for unreleased features.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [21.15.0.1] - 2025-12-28
+## [21.15.0.1] - Unreleased
 
 ### Added
 
--- a/gradle/build-logic/conventions/src/main/kotlin/SharedUtils.kt
+++ b/gradle/build-logic/conventions/src/main/kotlin/SharedUtils.kt
@@ -130,7 +130,7 @@ private fun Project.extractCurrentVersionChangelog(): String? {
 
     // Extracts the current version changelog without the version heading 2 and "For Devs" heading 3.
     val versionSectionRegex =
-        "(?s)## \\[$modVersion\\] - \\d{4}-\\d{2}-\\d{2}\\R(.*?)(?=\\R### For Devs|\\R## \\[.*?\\] |\\Z)"
+        "(?s)## \\[$modVersion\\] - (\\d{4}-\\d{2}-\\d{2}|Unreleased)\\R(.*?)(?=\\R### For Devs|\\R## \\[.*?\\] |\\Z)"
     val matcher = Regex(versionSectionRegex).find(fullChangelogText) ?: return null
 
     val versionChangelog = matcher.groupValues[1]


CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [21.15.0.1] - 2025-12-28
+## [21.15.0.1] - Unreleased
```
EchoEllet, line 8: I'm uncertain if this change is needed, since unreleased changes shouldn't even have a version, as shown in [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) (`## [Unreleased]`).

This way, it can be automatically released with a Gradle script with the correct date and version.

Typically, the version is changed depending on the new features or bug fixes.

Assuming the current version is `1.5.1`, the version componenets will be increased depending on the change (see also [SemVer](https://semver.org/)):

- one simple bug fix (`1.5.1` -> `1.5.2`)
- new feature such as emotes (`1.5.1` -> `1.6.0`)
- public API major breaking change or major mod rewrite (`1.5.1` -> `2.0.0`)

And as we add unreleased changes, the next release is still unknown, and therefore, we shouldn't decide the version earlier.

@Yesssssman 

I had a script that could eliminate these chores, and could send a PR if you would prefer to automate the workflow. [Original Dart script](https://github.com/singerdmx/flutter-quill/blob/master/scripts/publish_flutter_quill.dart), but it can be migrated to Gradle via Kotlin (`build.gradle.kts` or in `gradle/build-logic/conventions/src/main/SharedUtils.kt`).

This also uses a GitHub workflow to create diff links using Git tags (`v1.5.1`) automatically, and automate the changes of `gradle/versions.mods.toml` and `CHANGELOG.md` in one simple command.

EchoEllet (changes requested):

CHANGELOG.md
Yesssssman: The most reason I changed the date is patreon builds. They're not opened with changelog but still have a version with an alpha format.

And I still can't build the project with `## [Unreleased] - 2025-12-18` statement because the regular expression doesn't support it too.

> (?s)## \\[$modVersion\\] - (\\d{4}-\\d{2}-\\d{2}|Unreleased)\\r?\\n(.*?)(?=\\r?\\n### For Devs|\\r?\\n## \\[.*?\\] |\\Z)

Mod version should match with the changelog, which is specified under `libs.versions.toml`. It makes no sense we set the version as `Unreleased` because I still need versioning to publish patreon builds.

> I had a script that could eliminate these chores, and could send a PR if you would prefer to automate the workflow. [Original Dart script](https://github.com/singerdmx/flutter-quill/blob/master/scripts/publish_flutter_quill.dart), but it can be migrated to Gradle via Kotlin (build.gradle.kts or in gradle/build-logic/conventions/src/main/SharedUtils.kt).

I think the publishing workflow is automated enough. I want to build the project for both private exchange and patreon release the keep version consistency. (Often, when you make a change and share with others, it's better to change the version because two indifferent build can share the same version and re-distributed.)

EchoEllet: > The most reason I changed the date is patreon builds. 

We can use other alternatives, such as ignoring the automated publishing for Patreon builds, or replacing [the `error()`](https://github.com/Epic-Fight/epicfight/blob/7a414ee1b4a8d854989f66b9dc3e1fe152d42785/gradle/build-logic/conventions/src/main/kotlin/SharedUtils.kt#L183) with a warning message so you won't have to change this for Patreon builds.

> It makes no sense we set the version as Unreleased because I still need versioning to publish patreon builds.

This isn't the version; it's just a marker in `CHANGELOG.md` that will be replaced later.

> I think the publishing workflow is automated enough.

The Git tags are still a non-trivial feature that will be useful for comparing differences, porting, and catching regressions.
Being able to run `./gradlew publish -Pnew_version=1.0.0` is a nice bonus feature for easier maintenance.

EchoEllet: @Yesssssman 

Take a look at this commit: https://github.com/Epic-Fight/epicfight/commit/30ed89c49c8be5b5ebb5cfbb213328536afed11b

It will only produce a build failure when running the `publishMods` task and not the `build` task.

Yesssssman:
Closed cuz of better solution [here](https://github.com/Epic-Fight/epicfight/commit/30ed89c49c8be5b5ebb5cfbb213328536afed11b)

PR #2387, opened 2026-01-06
Disallow implementing EpicFightEventHooks or EpicFightClientEventHooks

EchoEllet:
Refactors `EpicFightEventHooks` and `EpicFightClientEventHooks` to use `public final class` with a private constructor and `public static final` for all fields.

This is **significantly** more verbose but **more Java idiomatic** since it disallows using `implements` for any of the interfaces.

Java's own classes already apply this pattern, and newer Fabric/NeoForge/Minecraft classes also adapt it:

image

image

image

It's more efficient to use patterns that prevent what shouldn't be done at compile time, to give a stronger semantics and clearer intent.

`EpicFightEventHooks` is only used as a structure or holder for these fields, and it's not meant to be subclassed/extended or implemented, or instantiated.

This also changes the IDE icon to be clearer:

<details><summary>Before</summary>
<p>

image

</p>
</details> 

<details><summary>After</summary>
<p>

image

</p>
</details>

### DIFF
--- a/neoforge/src/main/java/yesman/epicfight/api/client/event/EpicFightClientEventHooks.java
+++ b/neoforge/src/main/java/yesman/epicfight/api/client/event/EpicFightClientEventHooks.java
@@ -26,44 +26,69 @@ import yesman.epicfight.world.capabilities.entitypatch.LivingEntityPatch;
 /// only can register events that inherit [LivingEntityPatchEvent] for per entity events.
 ///
 /// For common event hooks for both client and server side, refer to [EpicFightEventHooks]
-public interface EpicFightClientEventHooks {
-    interface Camera {
-        CancelableEventHook<BuildCameraTransform.Pre> BUILD_TRANSFORM_PRE = CancelableEventHook.createCancelableEventHook();
-        EventHook<BuildCameraTransform.Post> BUILD_TRANSFORM_POST = EventHook.createEventHook();
-        EventHook<ItemUsedInDecoupledCamera> ITEM_USED_WHEN_DECOUPLED = EventHook.createEventHook();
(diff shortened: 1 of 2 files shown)


Yesssssman:
image

This is `net.minecraft.world.damagesource.DamageTypes` (minecraft original code). The clear benefit I see from this is we do not need `public static final` in front of each instance and no `private ClassName() {}` constructor.

I need to search more practices, also want to ask opinions from @Forixaim @sentwayfarer

Yesssssman:
image

One step elaborated approach from the raw `interface`. In this pattern, we can disallow implementing the data interface from outside of the file. It uses `sealed` interface featured in Java 15, pre-defining the classes that can implement this interface.

EchoEllet:
> This is `net.minecraft.world.damagesource.DamageTypes` (minecraft original code).

1. Minecraft codebase isn't meant to be extendable, and wasn't meant as a library for others to extend and build on top.
2. Minecraft codebase is based on legacy decisions from years ago and is consistently getting a lot of refactoring

E.g., now in 26.1, they are using `final` for every parameter, which wasn't the case in NeoForge 1.21.1:

image

IMo, inspirations should be taken from modern Java code over legacy Minecraft code, as they will catch up eventually, resulting in repeated refactoring.

image

>  The clear benefit I see from this is we do not need public static final in front of each instance and no private ClassName() {} constructor.
> One step elaborated approach from the raw interface. In this pattern, we can disallow implementing the data interface from outside of the file. It uses sealed interface featured in Java 15, pre-defining the classes that can implement this interface.

It comes with drawbacks and less obvious intentions. Using `interface` (even with `sealed`) as just a holder for the final fields. It's not a strong OOP design.

EchoEllet:
Closing since the project's code style is to use `sealed interface` over the proposed pattern.

PR #2382, opened 2026-01-05
Set up developer docs using WikiToolkit Gradle plugin

EchoEllet:
Setups mod developer docs using [Sinytra WikiToolkit](https://github.com/Sinytra/WikiToolkit) Gradle plugin within the same GitHub repository.

This uses the markdown files in `docs/`, and the publishing is fully automated, with no need for GitHub pages or full hosting setup.

Once this PR is merged, it will be available under [moddedmc.wiki/en/project/epicfight/latest](https://moddedmc.wiki/en/project/epicfight/latest).

Adding new guides or menus is straightforward.
Breaking changes or public API changes can be documented with the same commit that makes them.

This approach allows Epic Fight contributors to document public APIs and provide guides without needing to manage web-specific details such as HTML, CSS, JavaScript, hosting/domain, or deployment.

image

> [!NOTE]
> The docs' content will be fully refactored as we progress and rewrite the public APIs in the next major version.

### Preview

Checkout to this branch and run `./gradlew previewDocs` or watch this quick video:

<details><summary>Preview Video</summary>
<p>

https://github.com/user-attachments/assets/a4d6719c-f8eb-4d92-8aec-b54dbd77bf59

</p>
</details> 

### Justification

- A separate GitHub repo requires others to fork, create a branch, and commit for every change they make, which is time-consuming and inconvenient.
  -  This also requires more chores, such as fetching upstream changes and manual testing, whereas with this approach, it can be tested via `./gradlew previewDocs` with the same branch.
- The GitHub repository [Epic-Fight/EpicFight-Docs](https://github.com/Epic-Fight/EpicFight-Docs) is a full front-end project, containing numerous files, generated HTML pages, and too many bundled assets. It also includes ads and is hosted separately.
- The review process in the current WIKI is relatively slow and may take [a while](https://github.com/Epic-Fight/EpicFight-Docs/pulls?q=) for a PR. At the same time, other maintainers may want to introduce more changes or add more developer guides.
- More convenient to document breaking changes in `docs/migration/` within the same commit ([example file](https://github.com/singerdmx/flutter-quill/blob/master/doc/migration/10_to_11.md)) without two PRs, branches, commits, comments, or code reviews.

### [Existing Epic Fight WIKIs](https://epicfight-docs.readthedocs.io/)

The existing WIKIs can still be used for user guides, such as datapack, resource-pack, and so on.

PR #2390, opened 2026-01-07
Setting up automatic developer documentation syncing

guivnf:
Following up #2382, this Pull Request adds in a Github Workflow that automatically generates wiki pages inside [EpicFight-Docs](https://github.com/Epic-Fight/EpicFight-Docs). 

Any files added inside the ``root/docs/universal-docs/`` directory, will be automatically mirrored over to the EpicFight-Docs repository, which then automatically refreshes the [Official Wiki](https://epicfight-docs.readthedocs.io), serving the pages for the public's view.

***

Alternatively, I've also added an extra feature that allows us to create version specific documentation. Files added below ``root/docs/`` will automatically be reflected over to EpicFight-Docs too & will create a folder with the current branch name.

All of these pages will be displayed together inside the API tab:

image

Below is a simple comparison of the folder structure that's generated, to make it easier to understand what are universal documentation vs per version/branch documentation.

| Source Path (in EpicFight) | Target Path (in EpicFight-Docs) |
|---------------------------|--------------------------------|
| docs/universal-docs/General_Info.en.md | docs/API/General_Info.en.md |
| docs/Items.en.md (on branch 1.21.1) | docs/API/1.21.1/Items.en.md |
| docs/Items.en.md (on branch 1.20.1) | docs/API/1.20.1/Items.en.md |
| docs/Animations/Math.en.md (on branch 1.21.1) | docs/API/1.21.1/Animations/Math.en.md |

***

Finally, as a last addition to this PR, I've also updated the [Contributing Guide](https://github.com/Epic-Fight/epicfight/blob/1.21.1/CONTRIBUTING.md) so it's easier for new developers to understand how the documentation process and workflow works. 

This set of changes accomplishes the end-goal of automating the process of creating, editing, removing and publishing of API documentation pages, improving the workflow and making the overall process of sharing new additions to the API easier. 

As a last note, this was done through GitHub web, that's why the commit history is very dirty.


.github/workflows/mirror-docs.yml
```diff
@@ -0,0 +1,43 @@
+name: Mirror docs to EpicFight-Docs
+
+on:
+  push:
```
Yesssssman, line 4: Don't we also have to add `pull_request`? or it contains the case when PRs are merged to main branch?

Yesssssman:
This is unrelated, but is there a chance we can get out of the hosting website that posts ads? As we discussed before, I don't even want users to use some AdBlockers

image

guivnf:
> This is unrelated, but is there a chance we can get out of the hosting website that posts ads? As we discussed before, I don't even want users to use some AdBlockers
> 
> image

Sure, that can be arranged too. It'll take a bit more time since i'll need to move the entire website building ecosystem over to Github pages, but since it's already there it might be easier than what i expect.

.github/workflows/mirror-docs.yml
guivnf: Honestly, it was working whenever i did changes to files on that directory. One thing i couldn't manage was checking if the changes were made in the main branch, apparently Github doesn't have any ways to trigger the workflow if it was made on the default branch.
https://github.com/orgs/community/discussions/64921

sentwayfarer (approved):
LGTM.

sentwayfarer:
I do feel as if we should have a dedicated website for this though.

Yesssssman:
Let me test things internally. I'll create a new PR when I find further issues.

PR #2392, opened 2026-01-08
Removed unused skill build parameter

Yesssssman:
Removed the second parameter, which requires the skill class

Made this as PR since this is a breaking change for addons that have already ported to 1.21.1.


sentwayfarer (commented):
LGTM. Waiting for addon developers' approval.

PR #2393, opened 2026-01-08
Fixed mistype `DELIEVER_DAMAGE` -> `DELIVER_DAMAGE`

Yesssssman:
Fixed a mistype in event hook.

Made this as PR since this is a breaking change for addons that have already ported to 1.21.1.


sentwayfarer (commented):
Should probably wait for this one until we get confirmation from major addon developers. Good change on making these `public static final` though.

Yesssssman:
> Should probably wait for this one until we get confirmation from major addon developers. Good change on making these `public static final` though.

The skill API feature has not been published anywhere tho, I made this PR because ppl who can access EF repository might have already ported their mod to 1.21.1 and waiting for it to be released.

PR #2397, opened 2026-01-08
Added combo counter handler in CapabilityItem and fixed a crucial issue in combo attack

Yesssssman:
Feature requested by @Reascer, a unified and parameter-sensitive combo counter handler was required in development Weapons of Miracles.

The old API only provides a way to turn off combot counter reset feature when action animations are played.

```java
WeaponCapability.Builder builder = WeaponCapability.builder();
...
    builder.comboCancel(style -> false) // this completely turns off combo counter being reset when any action animation is played
...
```

Alternatively, the new API hands over various parameters that developers can use to handle the combo counter.
```java
WeaponCapability.Builder builder = WeaponCapability.builder();
...
    builder.comboCounterHandler((itemCapability, causal, entitypatch, nextAnimation, comboCounter) -> {
        // When causal is time expiration, reset the counter
        if (causal == ComboCounterHandleEvent.Causal.TIME_EXPIRED) {
            return 0;
        }
        
        List<AnimationAccessor<? extends AttackAnimation>> comboAnimations = itemCapability.getAutoAttackMotion(entitypatch);
        
        if (comboAnimations == null) {
            return 0;
        }
        
        Set<AnimationAccessor<? extends AttackAnimation>> attackMotionSet = Sets.newHashSet(comboAnimations);
        
        // when the next animation is not included in weapon attack motion, reset the counter
        if (!attackMotionSet.contains(nextAnimation)) {
            return 0;
        }
        
        int comboSize = comboAnimations.size();
        
        // When the next animation is dash or air attacks, reset combo counter
        if (nextAnimation.equals(comboAnimations.get(comboSize - 1)) || nextAnimation.equals(comboAnimations.get(comboSize - 2))) {
            return 0;
        }
        
        return comboCounter;
    })
...
```
By returning the modified combo counter, addons can select any combo animation in the sequence. This will open unlimited control of Basic Attack's combo counter.

### Fix for attack delaying when playing same animation from previous
This is a very simple fix, but the result was quite destructive so I leave the fix here.

Basic Attack uses an animation variable system where you can assign variables for specific animations. There is no problem in playing different animations sequentially, but it becomes a serious issue when playing the same one.

https://github.com/user-attachments/assets/7eef6863-0f09-4cfb-babf-563e6dc77dad

*note that I'm spamming attack key*

There is a significant delay until the next attack. This is more noticeable with animations that have longer recovery times especially heavy weapons in WoM. It was because of an unlogged exception, which never should be happened. You could see the exception by manually wrapping `try - catch`.

```
java.lang.UnsupportedOperationException: Can't modify a const variable
	at TRANSFORMER/epicfight@20.14.4/yesman.epicfight.api.animation.AnimationVariables.lambda$2(AnimationVariables.java:116)
	at java.base/java.util.HashMap.computeIfPresent(HashMap.java:1261)
	at TRANSFORMER/epicfight@20.14.4/yesman.epicfight.api.animation.AnimationVariables.put(AnimationVariables.java:112)
	at TRANSFORMER/epicfight@20.14.4/yesman.epicfight.api.animation.AnimationVariables.put(AnimationVariables.java:102)
...
```

This means you're trying to use an immutable animation variable because it detects the same variable key when the previous and next animations are the same.

The solution is swapping animation playing code and putting variable.

```diff
- executor.getAnimator().getVariables().put(COMBO, attackMotion, true);
- executor.getAnimator().playAnimation(attackMotion, 0.0F);
+ executor.getAnimator().playAnimation(attackMotion, 0.0F);
+ executor.getAnimator().getVariables().put(COMBO, attackMotion, true);
```

*Result:*

https://github.com/user-attachments/assets/3b60bfdd-c692-41ab-beb6-894219f4be08

You now see no delay between dash attacks.

(Additionally, I more like the bugged behavior, at least for spear😅 But I needed to correct misdesigned system)

### DIFF
--- a/src/main/java/yesman/epicfight/skill/BasicAttack.java
+++ b/src/main/java/yesman/epicfight/skill/BasicAttack.java
@@ -1,10 +1,10 @@
 package yesman.epicfight.skill;
 
 import java.util.List;
-import java.util.Set;
 import java.util.UUID;
 
-import com.google.common.collect.Sets;
+import org.jetbrains.annotations.Nullable;
+
 import net.minecraft.nbt.CompoundTag;
 import net.minecraft.network.FriendlyByteBuf;
 import net.minecraft.server.level.ServerPlayer;
@@ -18,7 +18,7 @@ import yesman.epicfight.api.animation.AnimationVariables.IndependentAnimationVar
 import yesman.epicfight.api.animation.property.AnimationProperty.ActionAnimationProperty;
 import yesman.epicfight.api.animation.types.AttackAnimation;
 import yesman.epicfight.api.animation.types.EntityState;
-import yesman.epicfight.api.animation.types.StaticAnimation;
+import yesman.epicfight.api.animation.types.MainFrameAnimation;
 import yesman.epicfight.network.EpicFightNetworkManager;
 import yesman.epicfight.network.common.AnimatorControlPacket;
 import yesman.epicfight.network.server.SPAnimatorControl;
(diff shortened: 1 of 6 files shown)


src/main/java/yesman/epicfight/skill/BasicAttack.java
```diff
@@ -1,10 +1,11 @@
 package yesman.epicfight.skill;
 
 import java.util.List;
-import java.util.Set;
 import java.util.UUID;
 
-import com.google.common.collect.Sets;
+import org.jetbrains.annotations.Nullable;
+
+import dev.tr7zw.skinlayers.versionless.util.Mth;
```
EchoEllet: This uses the `Mth` from 3D skin layers, which is an optional third-party mod dependency and may not be available at runtime.

Yesssssman: ooof thanks for pointing out

Issue #2364, opened 2026-01-02
More Balanced Dagger Weapon Innate


Lendoker:
### Details

I BEG you, please, at least add a damage limit for eviscerate on the dagger! It's really hard to balance bosses in the modpack when dagger trivializes ALL of them because of it's absurd damage output from the skill!

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #2417, opened 2026-01-14
[1.20.1] Added Eviscerate damage cap

Yesssssman:
Suggested by #2364

Applied a damage cap to the Eviscerate skill so it doesn't inflict massive damage against bosses with high health.

(Note the implementation is not good, but no breaking changes)


Yesssssman:
When this PR gets approved the change will be pushed to 1.21.1 too

Yesssssman:
## Attaching some testing images!

with damage cap

*With the damage cap 20 husk receives 25.8 damage (5.8 is from weapon damage)*


no damage cap

*Without the damage cap husk receives 133 damage*

sentwayfarer (approved):

Issue #2421, opened 2026-01-21
[Bug]: Phantom Ascent double jump breaks


Lendoker:
### Minecraft Version

1.20.1

### Epic Fight Version

21.14.6

### Minecraft Log Link

-

### Bug Details

Phantom Ascent doulbe jump fails to execute if a player didn't fall down 4 blocks

How to reproduce:
1. Have Phantom Ascent skill (at first it works normal)
2. Try to jump onto something, for example onto a two-block pillar using phantom ascent
3. Can't use Phantom Ascent anymore
4. Fall down from a 4 block height
5. Can use it again (??)

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Lendoker:
Additionally, when jumping for a second time even NOT in air you still perform Phantom Ascent and this bug triggers

Lendoker:
https://drive.google.com/file/d/1XOIU6tEr_DRUnUb2NX4l6yohyQGc0iTu/view?usp=sharing

Here's a link to the video showcasing the bug

Yesssssman:
Hey please download the temporary fix build and check if the issue where you can't use the skill again is solved. As I mentioned in #2425, I couldn't reproduce the bug, but I found some logically issue while trying to reproduce it. And I'm pretty sure it's related to that.

https://legacy.curseforge.com/minecraft/mc-mods/epic-fight-mod/files/7523478

Lendoker:
Yes, it is fixed, but in that version all items cause crushes (I mean ANY items, if any are dropped then it sends crushes), but overall, yes, the phantom ascent double jump issue is fixed

Yesssssman:
> Yes, it is fixed

Great, I'll mark this issue as resolved

> but in that version all items cause crushes

Please open a new tab with the crash report

PR #2425, opened 2026-01-25
[1.20.1] Fixed inconsistent "onFall" event call, causing phantom ascent unusable again

Yesssssman:
Currently, `LivingEntityPatch#onFall` relies on event hooks provided by Forge, but these events are only called when players **get hurt** by fall damage. This concept clashes with the design of Phantom Ascent, a mover skill that allows double jumping, which should reset the jump count when a player touches the ground, regardless of whether it gets damaged.

https://github.com/user-attachments/assets/00d20bd0-9615-4289-aa87-aad471fad2f4

When I put the output message to `onFall` event, it prints nothing.

image

onfall message insert

debug_output

So I searched for another hook that provides a reliable timing when a player touches ground, and I decided to mixin `Entity#setOnGroundWithKnownMovement` that is being called to set `Entity#onGround` field both in server and client.

Some tricks are applied to resolve the issue that Minecraft's API has in `MixinEntity`

**Demonstration**

https://github.com/user-attachments/assets/cb6dcdeb-38ec-4d26-9b4c-fcaac372ab07

debug_output_fixed

_The debug message now printed out properly_

Sadly, I wasn't able to reproduce the issue specified in #2421 (where you can't trigger phantom ascent again until you get hurt again), so I need to wait for the reporter's reply.

### DIFF
--- a/src/main/java/yesman/epicfight/events/EntityEvents.java
+++ b/src/main/java/yesman/epicfight/events/EntityEvents.java
@@ -36,12 +36,10 @@ import net.minecraftforge.event.entity.living.LivingDropsEvent;
 import net.minecraftforge.event.entity.living.LivingEquipmentChangeEvent;
 import net.minecraftforge.event.entity.living.LivingEvent;
 import net.minecraftforge.event.entity.living.LivingEvent.LivingJumpEvent;
-import net.minecraftforge.event.entity.living.LivingFallEvent;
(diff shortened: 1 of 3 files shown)


Yesssssman:
This PR will be merged automatically to 1.21.1 when all approved

sentwayfarer (approved):

PR #2511, opened 2026-03-29
Fixed Events replacing each other through priority systems matching.

Forixaim:
This PR fixes the underlying issue with ExCap and how the old Epic Fight's EventHook system works: each TreeMap<K, V> uses a single key and value, and event listeners that register at the same priority overwrite rather than append to the listener system.


Yesssssman:
LGTM. This should be applied to 26.1 build that I'm porting now ...

Yesssssman (approved):

Forixaim:
Alright, Starting Merge.

Yesssssman:
image

We need to use `CopyOnWriteArrayList` to avoid `ConcurrentModificationException` in Neoforge.

PR #2514, opened 2026-03-30
Use a ConcurrentHashMap instead of a TreeMap but still sort via priority.

Forixaim:
This is still a draft but feedback is still needed.

### DIFF
--- a/neoforge/src/main/java/yesman/epicfight/api/event/EventHook.java
+++ b/neoforge/src/main/java/yesman/epicfight/api/event/EventHook.java
@@ -7,9 +7,8 @@ import yesman.epicfight.api.utils.side.LogicalSide;
 
 import java.util.Comparator;
 import java.util.List;
-import java.util.TreeMap;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.stream.Stream;
 
 /// An event bus that is dedicated to one [Event] type
@@ -22,7 +21,7 @@ import java.util.stream.Stream;
 /// and [EpicFightClientEventHooks] for client-side only events
 public class EventHook<T extends Event> {
 	/// Treemap to order subscribers in descending order
-	final ConcurrentMap<Integer, List<EventListener<T>>> subscribers = new ConcurrentSkipListMap<>(Comparator.reverseOrder())
+	final ConcurrentMap<Integer, List<EventListener<T>>> subscribers = new ConcurrentSkipListMap<>(Comparator.reverseOrder());
 
     /// Determines if the event is only called in logical side
     protected final LogicalSide logicalSide;
@@ -33,13 +32,13 @@ public class EventHook<T extends Event> {
 
     /// Post the event to subscribers and execute tasks by their priority in descending order
     public T post(T event) {
-        List<EventListener<T>> buffer = Lists.newArrayList();
-        this.subscribers.values().forEach(buffer::addAll);
-        buffer.stream().sorted(Comparator.comparing(EventListener::priority)).forEach(listener -> {
-            if (listener.subscriptionType() instanceof DefaultEventSubscription<T> passiveSubscription) {
-                passiveSubscription.fire(event);
+        for (var subscriber : subscribers.values()) {
+            for (var listener : subscriber) {
+                if (listener.subscriptionType() instanceof DefaultEventSubscription<T> passiveSubscription) {
+                    passiveSubscription.fire(event);
+                }
             }
-        });
+        }
         return event;
     }
 
--- a/neoforge/src/main/java/yesman/epicfight/api/event/EventHook.java
+++ b/neoforge/src/main/java/yesman/epicfight/api/event/EventHook.java
@@ -22,7 +22,7 @@ import java.util.stream.Stream;
 /// and [EpicFightClientEventHooks] for client-side only events
 public class EventHook<T extends Event> {
 	/// Treemap to order subscribers in descending order
-	final ConcurrentMap<Integer, List<EventListener<T>>> subscribers = new ConcurrentHashMap<>();
+	final ConcurrentMap<Integer, List<EventListener<T>>> subscribers = new ConcurrentSkipListMap<>(Comparator.reverseOrder())
 
     /// Determines if the event is only called in logical side
     protected final LogicalSide logicalSide;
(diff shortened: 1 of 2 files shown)


Yesssssman:
There are compile errors should be fixed according to this change in `CancelableEventHook`

Yesssssman:
# Thread safety requirement
---
The subscriber container should be thread-safe since NeoForge loads mods asynchronously, if event registration is called in the mod's constructor it will cause `ConcurrentModificationException`.

Tested by custom multi threading code:

```java
public EpicFightMod(IEventBus modEventBus, ModContainer modContainer) {
        ...
        Thread t1 = new Thread(() -> {
            MutableInt mi = new MutableInt(0);

            for (int i = 0; i < 50; i++) {
                EpicFightEventHooks.Registry.ENTITY_PATCH.registerEvent(event -> {
                    System.out.println(mi.getAndIncrement() + "'th entity patch event");
                }, i);
            }
        });

        Thread t2 = new Thread(() -> {
            MutableInt mi = new MutableInt(1000);

            for (int i = 1000; i < 1050; i++) {
                EpicFightEventHooks.Registry.ENTITY_PATCH.registerEvent(event -> {
                    System.out.println(mi.getAndIncrement() + "'th entity patch event");
                }, i);
            }
        });

        Thread t3 = new Thread(() -> {
            MutableInt mi = new MutableInt(10000);

            for (int i = 10000; i < 10050; i++) {
                EpicFightEventHooks.Registry.ENTITY_PATCH.registerEvent(event -> {
                    System.out.println(mi.getAndIncrement() + "'th entity patch event");
                }, i);
            }
        });

        t3.start();
        t1.start();
        t2.start();
	}
```

```
Exception in thread "Thread-5" Exception in thread "Thread-4" java.util.ConcurrentModificationException
	at java.base/java.util.TreeMap.callMappingFunctionWithCheck(TreeMap.java:785)
	at java.base/java.util.TreeMap.computeIfAbsent(TreeMap.java:639)
```

*Output:*
```
10000'th entity patch event
10001'th entity patch event
10002'th entity patch event
10003'th entity patch event
1000'th entity patch event
1001'th entity patch event
1002'th entity patch event
1003'th entity patch event
1004'th entity patch event
1005'th entity patch event
1006'th entity patch event
1007'th entity patch event
1008'th entity patch event
1009'th entity patch event
1010'th entity patch event
1011'th entity patch event
1012'th entity patch event
1013'th entity patch event
1014'th entity patch event
1015'th entity patch event
1016'th entity patch event
1017'th entity patch event
1018'th entity patch event
1019'th entity patch event
1020'th entity patch event
1021'th entity patch event
1022'th entity patch event
1023'th entity patch event
1024'th entity patch event
1025'th entity patch event
1026'th entity patch event
1027'th entity patch event
1028'th entity patch event
1029'th entity patch event
1030'th entity patch event
1031'th entity patch event
0'th entity patch event
1'th entity patch event
2'th entity patch event
3'th entity patch event
4'th entity patch event
5'th entity patch event
6'th entity patch event
7'th entity patch event
8'th entity patch event
```

Due to the registration failing while running the thread, the event logs end in a random sequence.

# Performance Optimization
___
Instead of sorting event priorities each time an event hook is called, we should pre-order the events so that the event hook just calls subscribers in order.

Tested code on PR (sort on trigger) vs local test code (sort on register)

```java
public T post(T event) {
    long begin = System.currentTimeMillis();

    for (var subscriberList : this.subscribers.values()) {
        for (var subscriber : subscriberList) {
            if (subscriber.subscriptionType() instanceof DefaultEventSubscription<T> passiveSubscription) {
                passiveSubscription.fire(event);
            }
        }
    }

    long took = System.currentTimeMillis() - begin;
    System.out.println("ConcurrentSkipListMap + CopyOnWriteArrayList event call: " + (took));

    long begin2 = System.currentTimeMillis();
    List<EventListener<T>> buffer = Lists.newArrayList();

    this.subscribers.values().forEach(buffer::addAll);
    Comparator<EventListener<T>> eventListeners = Comparator.comparing(EventListener::priority);

    buffer.stream().sorted(eventListeners.reversed()).forEach(listener -> {
        if (listener.subscriptionType() instanceof DefaultEventSubscription<T> passiveSubscription) {
            passiveSubscription.fire(event);
        }
    });

    long took2 = System.currentTimeMillis() - begin2;

    System.out.println("concurrent map + arraylist event call: " + (took2));


    return event;
}
```

*Output*
```java
ConcurrentSkipListMap + CopyOnWriteArrayList event call: 187
concurrent map + arraylist event call: 690
```

p.s. The sort order of event subscribers is descending

neoforge/src/main/java/yesman/epicfight/api/event/EventHook.java
```diff
@@ -20,7 +22,7 @@
 /// and [EpicFightClientEventHooks] for client-side only events
 public class EventHook<T extends Event> {
 	/// Treemap to order subscribers in descending order
-	final TreeMap<Integer, List<EventListener<T>>> subscribers = new TreeMap<> ((i1, i2) -> Integer.compare(i2, i1));
+	final ConcurrentMap<Integer, List<EventListener<T>>> subscribers = new ConcurrentHashMap<>();
```
Yesssssman:
```suggestion
	final ConcurrentMap<Integer, List<EventListener<T>>> subscribers = new ConcurrentSkipListMap<>(Comparator.reverseOrder())
```

`ConcurrentSkipListMap` will ensure sorting listeners in priority on registration (works same as `TreeMap`)

neoforge/src/main/java/yesman/epicfight/api/event/EventHook.java
```diff
@@ -31,14 +33,13 @@ protected EventHook(LogicalSide logicalSide) {
 
     /// Post the event to subscribers and execute tasks by their priority in descending order
     public T post(T event) {
-
```
Yesssssman, line 34: We can simply go back to the old code to trigger events with no extra sorting process.

neoforge/src/main/java/yesman/epicfight/api/event/EventHook.java
Yesssssman:
```diff
-		this.subscribers.computeIfAbsent(priority, sub -> Lists.newArrayList()).add(new EventListener<>(name,priority,subscription));
+		this.subscribers.computeIfAbsent(priority, sub -> new CopyOnWriteArrayList<> ()).add(new EventListener<>(name, priority, subscription));
```

`CopyOnWriteArrayList` will ensure thread-safety in registering same priority events

Yesssssman (changes requested):

neoforge/src/main/java/yesman/epicfight/api/event/EventHook.java
Forixaim: Done

neoforge/src/main/java/yesman/epicfight/api/event/EventHook.java
Forixaim: Done

Yesssssman (approved):

PR #2521, opened 2026-04-05
Added Collider scaling based on InteractionRange Attribute

### LINKED ISSUE #2186: [Suggestion]: Global/specific attribute/config for weapon reach
### Have you checked if there's any similar suggestions ?

- [x] I checked, there are no similar suggestions that have been reported.

### Are you using the latest Epic Fight ?

- [x] I checked I'm using latest Epic Fight version.

### Is this suggestion related to another mod? (i.e: add animations for X mod)

- [ ] This is related to other mods.

### The mod

_No response_

### Minecraft Version

1.20.1

### What's your suggestion?

Weapon reach can be very incongruent with the weapon type being used, such as with spears, halberds, or most 2 handed weapons. This leaves the weapons sometimes missing or requiring the player to be too close to the enemy. I've tried upping the entity_reach forge attribute, but it seems that epic fight doesn't apply this. I assume that epic fight is using the weapon model itself to determine a hit, much like Better Combat. But Better Combat has an attribute/config to allow for reach modification. 

My suggestion is to implement a /attribute, such as epicfight:reach that can be modified or implement a config option that allows a global reach value to be entered and specific items to have their reach modified. Thank you.

JacobFrodermann:
This PR aims to add weapon collider scaling base on the ENTITY_INTERACTION_RANGE attribute, this would improve compatibility with mods providing a reach buff.
The scaling is calculated based on the base interaction range. 

Would resolve #2186

### DIFF
--- a/neoforge/src/main/java/yesman/epicfight/api/animation/types/AttackAnimation.java
+++ b/neoforge/src/main/java/yesman/epicfight/api/animation/types/AttackAnimation.java
@@ -29,6 +29,7 @@ import yesman.epicfight.api.animation.property.MoveCoordFunctions;
 import yesman.epicfight.api.animation.types.EntityState.StateFactor;
 import yesman.epicfight.api.asset.AssetAccessor;
 import yesman.epicfight.api.collider.Collider;
+import yesman.epicfight.api.collider.WeaponReachHitboxUtil;
 import yesman.epicfight.api.event.EpicFightEventHooks;
 import yesman.epicfight.api.event.types.animation.AttackPhaseEndEvent;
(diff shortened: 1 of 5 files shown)


Yesssssman (commented):
Could you also consider calculating `WeaponCapability#reach` property? That feature distances an attacker from a target for `ComboAttack` playing.

If the reach property is bigger than collider Z length you can multiply the box as `ENTITY_INTERACTION_RANGE` works.

Forixaim (commented):
LGTM but yeah yesman's got a point.

JacobFrodermann:
yeah i am going to implement it

PR #2527, opened 2026-04-08
retextured weapons and adjusted their hitboxes

Forixaim:
This update readjusts some weapon textures to be more in-line with how they are depicted in real life as well as keeping weapons pixel consistent with each other.


Yesssssman (approved):
LGTM

PR #2532, opened 2026-04-10
Add compatibility with Wildfire's Female Gender mod + ExCap server sided bug fixes

Forixaim:
This PR adds compatibility with Wildfire's Female Gender Mod, which currently cancels the render and defers it to the patched render, using clever matrix transformations to make it work.

Additionally it also addresses issue #2533


Yesssssman (approved):
LGTM. Maybe implementing NetSyncListener for existing data pack reloaders will ensure better consistency

PR #2537, opened 2026-04-11
fix compute shader render error while persistent buffer turn off.

dfdyz:
bug example
image
Although this issue has been fixed, the frame rate drops significantly when a large number of entities are displayed on the same screen. 
It is recommended that players update their graphics card drivers(which support OpenGL4.6) and then enable the persistent mapping buffer.


dfdyz:
I know what happen. The fg gradle modifiy my jdk automaticlly, and that jdk not be forced to use the NV gpu.

dfdyz:
I will add a function to detect the current OpenGL runtime platform in the future, reminding players to use a discrete graphics card or disable computational shaders.

Yesssssman:
> I know what happen. The fg gradle modifiy my jdk automaticlly, and that jdk not be forced to use the NV gpu.

Which JDK version was that? Does that only happen in the dev environment?

Forixaim:
> I know what happen. The fg gradle modifiy my jdk automaticlly, and that jdk not be forced to use the NV gpu.

For 1.20.1 I plan give one last overhaul on the gradle script by switching to ModDevGradle + Kotlin DSL so no worries i guess.

Forixaim (approved):
LGTM as well.

Yesssssman (approved):
I'll do more test in my local

PR #2543, opened 2026-04-15
[1.21.1] Fixed shields not rendering offhand

Forixaim:
A small fix that makes shields render offhand when alone


Yesssssman (approved):
LGTM, I'll just merge cuz it's a simple change

PR #2547, opened 2026-04-19
Use @WrapOperation instead of @Redirect 

Nikdo53:
Even though they they're fine by themselves, redirects don't stack with other mods. This is an incompatibility just waiting to happen and it will be hard to debug once it does. 

Since your use of redirects is exactly what [@WrapOperation from mixin extras does](https://github.com/LlamaLad7/MixinExtras/wiki/WrapOperation), I've used it in their place. Everything should be the same with the except MixinGui.epicfight$renderCrosshair where I added runnable for the original crosshair (since its called from an inject too)


Forixaim:
I'll run gameplay tests once I'm home from work.

Forixaim (approved):
LGTM

Forixaim:
PR has been moved to #2565 I sincerely thank you for pointing it out.

PR #2553, opened 2026-04-26
Improving onboarding issues by simplifying registration.

### LINKED ISSUE #2554: [1.21.1] Switch from AnimationAccessor into a Hybrid Registry using DeferredRegister + ReloadListener
### Use case

Problems with the old Animation API
It involved being an event-driven way to register animations via code which is extremely tedious and if someone forgets to hook onto the event their entire animation doesn't load.

### Proposal

Switch how you register Animations from hooking into AnimationRegistryEvent event into DeferredRegister
It's easier, quicker, less mistakes, and is much better at fixing static issues regarding AnimationAccessors being null.

It would also retire AnimationAccessor as we fully embrace vanilla minecraft's Holder.

Forixaim:
This PR improves onboarding issues by significantly simplifying how a developer registers their weapon with ExCap this gets rid of all the complicated events and just uses one static register for each value.


Forixaim:
# Some issues I've came across that can be solved by likely moving animations to DeferredRegister:

## Null accessors
- The event driven architecture meant that when passing Animations into a Moveset gave me a list of nulls.
- Caused by Movesets picking up the animations while it was still null and keeping the null when animations got updated.
 
## Proposed Solution: Move Animation Registration into a Hybrid Registry System
- Code defined animations should likely stay statically declared. It's how Minecraft does with its items and such.
- This involves using a DeferredRegister<> for animations instead of using our old event-driven registry.
- Yes this would cause a lot of issues and will break the public facing API, so here's the benefits:
  - It's a lot easier to register animations by code, and I have plans to make it as easy as registering an item.
    - For example registering an animation the old way would take around 5-6 steps, this new way would take 3 steps.
    - Keeping the hybridized architecture means that the registry map would then be appended into the runtime map where datapacks can override if needed.
    - Eliminates the need for an event-driven registry system which is very legacy forge things.
    
## Existing Changes
- Moved ExCap and I would say registration into utilizing a Hybridized DeferredRegister + Reload Listener since it's statically safe, helps error checking duplicate builders, and is quick and easy. The hybrid register registers code based capabilities, modifiers, and such statically via DeferredRegister which is then placed into a Reload Listener's runtime map, where datapacks can override.
- The hybrid register was a concept created from the backlash and onboarding issues with the old ExCap implementation. This architectural change gives more compatibility with KubeJS while keeping Datapacks included.
 
Since this is a public API breaking change, I'd propose slowly rolling out betas for addon developers to experiment around and adapt their code. That way if they have objections or any feedback we can quickly go in and patch and release a new beta. 

Edit: Disregard

Forixaim:
# Modifiers are the new way to modify Epic Fight weapons dynamically

## Additionally, ExCap weapons can be registered monolithically, with a single builder class holding the weapon attributes, movesets, and style conditionals, or, for scalability and reusability, conditionals and movesets can be registered modularly as separate components that can be added. For example: A player wants to add a new moveset to Bokken that works like a Longsword. But oops! The builder is inaccessible and must be modified. Never worry: by registering the modifier and targeting that specific weapon, players can modify it with a single static initialization and definition of said modifier.

Forixaim:
Fixed issue where Animations were null, DeferredRegister can continue

guivnf (changes requested):
1. Restore the 5 events as ``@Deprecated`` shims that fire alongside the new managers; remove in a fuuture major. Removing 5 public events with no deprecation cycle = a literal API breach, the exact thing the addon devs asked for ("stable API") is what this violates. 
the new managers (ConditionalManager, ItemPresetManager, etc.) are annotated @ApiStatus.Experimental / @ApiStatus.Internal. So this PR removes a stable public API and replaces it with one explicitly marked experimental

2. A new yesman.epicfight.EpicFight class is introduced alongside the existing yesman.epicfight.main.EpicFightMod, and 6+ logger sites switch from EpicFightMod.LOGGER to EpicFight.LOGGER. Why two parallel main/logger classes? Either consolidate, or document the intended split leaving both around invites addons to pick the wrong one and silently desyn c

3. Hybrid ``DeferredRegister ``+ runtime override map's precedence is unspecified, you should document precedence rules for datapack vs. ``DeferredRegister`` and for multimodifier conflicts; add a priority/order field to ``WeaponModifier``

4. Record components are List<ResourceLocation>, Map<…>, Map<…>. Records only enforced reference immutability, if the manager hands the same maps out to consumers, mutation leaks. Confirm Builder.build() wraps with List.copyOf / Map.copyOf

5. Holder<SoundEvent> / Holder<ParticleType<?>> migration
the existing public builder method .hitSound(SoundEvent) needs either to keep an SoundEvent overload or every addon's .hitSound(EpicFightSounds.BLADE_HIT.get()) call breaks. Worth confirming both shapes compile.

6. The CHANGELOG should clearly say "ExCap registration moved to DeferredRegister; Modifiers are an additional dynamic‑override layer", they're not synonyms

7. No mention of tests, sample addons, or migration guide, i'd consider trying some kind of addon along side these changes.

8. Replacing event posting with a direct Manager.acceptEvent() call is fine internally but this is also where 3rd‑party hooks lived. If the only hook surface left is "register a DeferredHolder," there's no lat e/dynamic way for compatibility mods (e.g., compatibility mods that previously subscribed to these events for late/dynamic reactions) to react to other adddons' registrations. the old event bus gave them that for free i think

Please commit to these changes as i think they are vital for the stability of this PR & API comming forward

Forixaim:
Tested with Rapier Moveset, Dual Spears, Dual Axes, and Bow Moveset (There aren't a lot of addons on 1.21.1)
No regressions as of now.

guivnf (approved):
Looks great, well done, thanks for the changes!

PR #2555, opened 2026-04-28
Revert ExCap changes while retaining compatibility

Forixaim:
This reverts changes due to ExCap being a sudden API shift out of nowhere.


guivnf (changes requested):
Consider bumping to ``20.14.17`` or ``20.14.16.2`` instead. The CHANGELOG section for 20.14.16.1 is removed. Reverting code is fine, deleting prior changelog history makes it harder for users to understand what happened. A "``20.14.17`` reverted ExCap" entry that acknowledges ``20.14.16.1`` existed would be cleaner than pretending it didn't.

the deprecation annotation tightening (``@Deprecated(since = "1.21.1", forRemoval = true)``) added by ``0b8b426e`` is being reverted along with the dodge getters, since they were in the same commit. That's a slight loss those annotations were independently useful.

minhhjj5678:
@Forixaim  Will the ExCap API be backported to 1.20.1 again in the future? I spent a lot of time making my mod compatible with this new API.

guivnf:
> @Forixaim Will the ExCap API be backported to 1.20.1 again in the future? I spent a lot of time making my mod compatible with this new API.

No it won't, we had that API first be implemented in 1.21.1, then backported to 1.20.1, however, other developers disliked the changed and so we reverted it. It's no longer necessary and the process of making your addon should be simpler now. If you need help with anything feel free to ask on our discord, if you don't have the developer role already, simply ask a moderator to gain access to the developer help channels.

PR #2557, opened 2026-04-30
Fix AzureLib 3.x compatibility + drop multiloader scaffolding

guivnf:
This PR does two unrelated cleanups that both fell out of getting the mod building cleanly again on 1.21.1. 

## 1. AzureLib / AzureLib-Armor 3.x compat
Problem: Newer AzureLib (``3.x``) reorganised its rendering pipeline. The classes our compat code targeted (`GeoArmorRenderer` `RenderProvider`, `GeoBone`, `BoneSnapshot`, `mod.azure.azurelib.common.internal.client.util.RenderUtils`) were either removed or moved. Items also stopped registering through `IClientItemExtensions.getHumanoidArmorModel` and now use `AzArmorRendererRegistry` instead. Net result on a user-side install: AzureLib armors rendered with the missing-texture pink/black checker pattern and the wrong (vanilla) geometry.

To make matters worse, there was a hard-coded gate in `EpicFightMod.getCompatibilityModules` that explicitly disabled `AZURE_LIB` / `AZURE_LIB_ARMOR` whenever the installed AzureLib was >= 3.0.0, so the compat module was never even loaded.

## 2. Drop the multiloader scaffolding
The repo carried `:common` and `:fabric` Gradle subprojects plus `multiloader-base` / `multiloader-platform` convention plugins. There hasn't been an actual Fabric build of this mod for a long time , `:common `only held 7 platform-abstraction .java files and the resource bundle, `:fabric` held five stubs. Past 1.x branches (e.g. 1.20.1) were single-loader projects with src/ directly under the repo root, so this brings the layout back in line with that history.

## Misc
Deleted `api/utils/TemporaryCode.java` (empty class with only a commented out body, zero references — a dead `@Deprecated` artifact).

### Observation:
image
The azurelib compat was tested against Samurai Dynasty mod, & works just fine as shown on the image above.


Forixaim (approved):
LGTM

PR #2562, opened 2026-05-11
[21.17.0] Animation Mirroring / QOL Changes

guivnf:
## Summary

Reworks how the mod handles offhand-side animations. Previously the offhand variants of drink/eat/spyglass/shield/dig were collapsed into a single shared animation and the offhand-wielding case for weapons had no rendering support at all a sword in the offhand with an empty mainhand rendered as if the arm were empty. This PR introduces a proper mirror system,  and fixes the follow-on issues that surfaced once both paths existed at the same time.

### `PoseMirror` utility

New helper that reflects a `Pose` across X = 0: negate translation X, conjugate the rotation quaternion (`qy`/`qz` negated), swap `*_R`/`*_L` joint names, rebuild the per-joint `TransformEntry` map so mask-driven auxiliary transforms also mirror. Two-arg variant offers `preserveRoot` because the per-layer wrapper and the global mirror-mode pass want different policies on `Root` (KEEP_CHILD_LOCROT compensation vs. authored throw lean). FrontResult preservation policy is per-caller for the same reason camera-tracking yaw stays world-aligned while authored body lean reflects normally. 

### `MirroredAnimation` / `DirectMirroredAnimation`

Pose-time decorator that wraps a source `StaticAnimation` and reflects its output. Used for offhand equip/unequip flourishes and the offhand dig clip a single authored source covers both hands, the wrapper gives the offhand variant its `left_arms` mask via auto-flipping `JointMaskEntry::mirror`. `hasTransformFor` flips the queried joint name back so `ClientAnimator#applyBindModifier`'s `Root` KEEP_CHILD_LOCROT hook fires for the mirrored joints.

### `MirrorAnimation` (two-path) KEPT, with smarter hand routing

Kept the explicit-file approach for animations that have authored per-hand clips. The constructor takes a `mainhandPath` and an `offhandPath`; both load as `DirectStaticAnimation`. Animations that still use it for reference:

  - `BIPED_DRINK`, `BIPED_EAT`, `BIPED_SPYGLASS_USE`
  - `BIPED_BLOCK` (shield raise), `BIPED_HIT_SHIELD` (on-hit reaction)
  - `BIPED_HOLD_MAP_MAINHAND` / `_OFFHAND` (+ `_MOVE` variants)
  - `BIPED_DIG` (via `SelectiveAnimation` over `dig_mainhand`/`dig_offhand`)
  
## QOL Changes & Tweaks/Fixes
- Fixed biped entities mounted atop horses & other entities rendering all weird. The skeletons on top of spiders (spider jockeys) should be more visually pleasing. Same goes for the Chicken jockey.
- Fixed the tools being misplaced on skeletons when they sit/mount on other entities
- Fixed chicken jockey behaviour, now it attacks the player and targets it properly.
- Added Hand switching item animation
- Added max passive skills gamerule. `/gamerule maxPassiveSkills <value>`
- Made it so roll/dodge skills cannot be performed mid air.
- Made it so you cannot spam dash attacks. 
- Tweaks to the Skill equipping window GUI. No longer closes when equipping a skill and some other tweaks here and there.

## Testing
image
Tested against "Rapier Mod" addon. Seems to be working perfectly.


Forixaim (approved):
Looks good.

PR #2565, opened 2026-05-14
[1.21.1] 21.17.2 Release, Improving Compatibility with Mixins by using @WrapOperation instead of @Redirect, and More Addon Control

Forixaim:
Used WrapOperation instead of Redirect for better compatibility. Following PR #2547. A small API addition that gives explicit item overrides for those who craved the one bit of extra control over WeaponCapabilityPresets


guivnf:
LGTM

